PR #22, opened 2012-11-19
Ignore test files even if they are not in the top level

cyli:
Currently, pep258 only ignores files if they are in the same directory in which pep257.  This ignores all test files in subdirectories as well.

### DIFF
--- a/pep257.py
+++ b/pep257.py
@@ -65,6 +65,7 @@ Also, see examples in "Check functions" section.
 
 """
 
+import os.path
 import re
 import inspect
 from curses.ascii import isascii
@@ -329,7 +330,8 @@ def check_source(source, filename):
     keywords = ['module_docstring', 'function_docstring',
                 'class_docstring', 'method_docstring',
                 'def_docstring', 'docstring']  # TODO? 'nested_docstring']
-    is_script = source.startswith('#!') or filename.startswith('test_')
+    basename = os.path.basename(filename)
+    is_script = source.startswith('#!') or basename.startswith('test_')
     for keyword in keywords:
         for check in find_checks(keyword):
             for context in parse_contexts(source, keyword):
--- a/test_pep257.py
+++ b/test_pep257.py
@@ -238,3 +238,25 @@ def test_opened_files_are_closed():
     assert len(files_opened) == 1
     for opened_file in files_opened:
         opened_file.close.assert_called_once_with()
+
+
+def test_ignore_scripts():
+    import mock
+
+    def check_is_script(doc_string, context, is_script):
+        assert is_script
+        return None
+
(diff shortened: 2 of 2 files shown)


jayvdb:
It would be good if pep257 supported the same command line (and config) options as pep8:

    --exclude=patterns   exclude files or directories which match these comma
                         separated patterns (default:
                         .svn,CVS,.bzr,.hg,.git,__pycache__)

keleshev:
@jayvdb Now `pep257` has `--match` and `--match-dir`:

```
--match=<pattern>     check only files that exactly match <pattern> regular
                        expression; default is --match='(?!test_).*\.py' which
                        matches files that don't start with 'test_' but end
                        with '.py'
--match-dir=<pattern>
                        search only dirs that exactly match <pattern> regular
                        expression; default is --match-dir='[^\.].*', which
                        matches all dirs that don't start with a dot
```

Which are designed to cover all the discussed use-cases.

jayvdb:
@keleshev , that isn't very nice if we want to exclude a single file, or a few dirs, and we need to maintain exclusion rules for pep8 & flake8.

e.g.
http://git.wikimedia.org/blob/pywikibot%2Fcore.git/b83e2e9e7dbf01800cc4b3bdef7f31d8ee775fdf/tox.ini#L139

Ideally, projects can add a pep257 block that is identical to pep8 and flake8 to their tox.ini.  Having to maintain a second, more complex inverse, regex to do the same for pep257 alienates people who are not regex gurus.

sigmavirus24:
@jayvdb you realize you don't need to bother if you just use flake8-docstrings, right?

jayvdb:
@sigmavirus24 , sure am, and we are doing that (see linked tox.ini).  We were doing that on patch upload, but that has now changed so that our linters can only run after patch approval (+1).

There are a lot less moving parts if we can run pep8 and pep257 directly, and from a security policy perspective there is less to audit if linters are run without a generic and extensible frameworks like flake8 & tox.

sigmavirus24:
> There are a lot less moving parts if we can run pep8 and pep257 directly,

Fair enough.

> from a security policy perspective there is less to audit if linters are run without a generic and extensible frameworks like flake8 & tox.

This doesn't make much sense though and is probably better suited for discussion on code-quality@python.org or ##python-code-quality on Freenode.

jayvdb:
Would you be happy with 'exclude' being added?

If yes, ... some details that are unimportant to me, but may as well decide before coding: 
The same order of precedence as pep8 wrt --match/match-dir?
As a glob or a regex?

samj1912:
This PR has not had any updates in 5 years. Closing this for now. If there's any interest in resuming this, feel free to reopen.

PR #69, opened 2014-04-14
Make setters private

adiroiban:
Problem
---------

pep257 complains that setters (and mabye deleters) have no docstring.

Here is an example

```
    @property
    def enabled(self):
        """
        See: `IEntityConfiguration`.
        """
        return self._proxy.getBoolean(self._section_name, 'enabled')

    @enabled.setter
    def enabled(self, value):
        return self._proxy.setBoolean(
            self._section_name, 'enabled', value)
```

Documentation tools should ignore setters and deletes docstring when creating api docs.
Also, setters and deleters are pretty standard and they will only have a boilerplate text which will not say anything new.

Changes
------------

Updated the parser to keep a record for decorators.

In Method.is_private i check the list of decorators and if decorators start with same name then it is marked private.

I have removed unused / broken tests from test_pep257 as this is just hoarding of junk. http://blog.8thlight.com/uncle-bob/2014/04/03/Code-Hoarders.html

Added test for new  code.

How to test
----------------

Check that changed make sense

```py.test test_pep257.py -v```

Let me know if you find this useful.

Thanks!

### DIFF
--- a/pep257.py
+++ b/pep257.py
@@ -66,7 +66,8 @@ class Value(object):
 
 class Definition(Value):
 
-    _fields = 'name _source start end docstring children parent decorators'.split()
+    _fields = (
+        'name _source start end docstring children parent decorators'.split())
 
     _human = property(lambda self: humanize(type(self).__name__))
     kind = property(lambda self: self._human.split()[-1])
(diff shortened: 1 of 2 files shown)


glennmatthews:
This can probably be closed now that #107 is merged.

PR #75, opened 2014-07-01
Parse arguments using docopt

Nurdok:


sigmavirus24:
I wonder if using docopt will limit (and potentially break) the flake8-docstrings extension. We don't currently muck with the options provided by pep257 but we will not that the tool provides a `--count` flag.

keleshev:
Believe it or not, I'm not a big fan of this change either. I liked having no dependencies.

Nurdok:
@sigmavirus24, I didn't quite understand what you're saying. In what way will flake8-docstrings will be limited by this change?
@keleshev, I know, I'm hesitant about adding dependencies as well and that's why I haven't merged this PR yet. Although I'm not sure if having no dependencies is such a big advantage anyway - I certainly wouldn't mind this when using pep257 in my projects.

Issue #86, opened 2014-12-31
Add a `--count` flag that prints the number of violations found


Nurdok:
This is useful to get a metric for code conformance.
It will also help marking progress when fixing a large code base with many violations.

PR #89, opened 2015-01-03
Added `--count` flag.

Nurdok:
As per #86.

### DIFF
--- a/pep257.py
+++ b/pep257.py
@@ -412,7 +412,7 @@ def get_option_parser():
     parser = OptionParser(version=__version__,
                           usage='Usage: pep257 [options] [<file|dir>...]')
     parser.config_options = ('explain', 'source', 'ignore', 'match',
-                             'match-dir', 'debug', 'verbose')
+                             'match-dir', 'debug', 'verbose', 'count')
     option = parser.add_option
     option('-e', '--explain', action='store_true',
            help='show explanation of each error')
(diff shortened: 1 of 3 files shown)


pep257.py
```diff
@@ -564,10 +566,15 @@ def main():
...
+    errors = check(collected, ignore=options.ignore.split(','))
     code = 0
-    for error in check(collected, ignore=options.ignore.split(',')):
+    count = 0
+    for error in errors:
         sys.stderr.write('%s\n' % error)
         code = 1
+        count += 1
+    if options.count:
+        print(count)
```
sigmavirus24: Why not just `print(len(errors))` at the end instead of keeping a count variable?

Nurdok: Because `errors` is a generator object. I also think this is a bit ugly. I'd love to hear suggestions.

sigmavirus24: Ah, I forgot `errors` was a generator. This is fine then.

Nurdok: In any case, I really appreciate the effort of going through pull requests. Thanks @sigmavirus24!

sigmavirus24: Always happy to help. As maintainer of flake8-docstrings I think I'll have to account for this new flag in the next release because `--count` is a flag already offered by pep8. (It's what piqued my interest particularly. :))

keleshev: How about:

```python
    errors = check(collected, ignore=options.ignore.split(','))
    count = 0
    for count, error in enumerate(errors, 1):
        sys.stderr.write('%s\n' % error)
    if options.count:
        print(count)
    return bool(count != 0)
```

sigmavirus24: It's simpler, but I'm not sure it's that important of an optimization.

PR #103, opened 2015-02-08
Use function definitions instead of lambdas

### LINKED ISSUE #102: Fix pep8 violations according to new 1.6.0 version
There was a new release of pep8 yesterday (Feb 6th, 2015) that added some new checks that pep257 violates (most notably lambda expression assignment). We need to fix these violations for our build to pass.

sigmavirus24:
Fixes #102

### DIFF
--- a/pep257.py
+++ b/pep257.py
@@ -87,10 +87,8 @@ class Value(object):
         return other and vars(self) == vars(other)
 
     def __repr__(self):
-        kwargs = ', '.join(
-            map(lambda arg: '{}={!r}'.format(arg, getattr(self, arg)),
-                self._fields)
-            )
+        kwargs = ', '.join('{}={!r}'.format(field, getattr(self, field))
+                           for field in self._fields)
         return '{}({})'.format(self.__class__.__name__, kwargs)
 
 
--- a/pep257.py
+++ b/pep257.py
@@ -54,22 +54,43 @@ __all__ = ('check', 'collect')
 
 PROJECT_CONFIG = ('setup.cfg', 'tox.ini', '.pep257')
 
-humanize = lambda string: re(r'(.)([A-Z]+)').sub(r'\1 \2', string).lower()
-is_magic = lambda name: name.startswith('__') and name.endswith('__')
-is_ascii = lambda string: all(ord(char) < 128 for char in string)
-is_blank = lambda string: not string.strip()
-leading_space = lambda string: re('\s*').match(string).group()
+
+def humanize(string):
+    return re(r'(.)([A-Z]+)').sub(r'\1 \2', string).lower()
+
+
+def is_magic(name):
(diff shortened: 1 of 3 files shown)


sigmavirus24:
So it looks as if whatever plugin you're using to run pep8 is doing the equivalent of

```
$ pep8 --ignore=E701
```

But `E704` is off by default. Should we add `E704` to the ignore list as well?

pep257.py
```diff
@@ -54,22 +54,43 @@ def next(obj, default=nothing):
...
+    def __eq__(self, other):
+        return other and vars(self) == vars(other)
 
     def __repr__(self):
-        format_arg = lambda arg: '{}={!r}'.format(arg, getattr(self, arg))
-        kwargs = ', '.join(format_arg(arg) for arg in self._fields)
+        kwargs = ', '.join(
+            map(lambda arg: '{}={!r}'.format(arg, getattr(self, arg)),
+                self._fields)
+            )
```
Nurdok: I think this change just makes the code both longer and less readable.

Nurdok:
I left one comment. Otherwise, this is great. Thank you so much for your time and effort!

pep257.py
jirikuncar: What about something like:
```python
kwargs = ','.join('{}={!r}'.format(arg, getattr(self, arg)) for arg in self._fields)
```

pep257.py
```diff
@@ -100,7 +123,9 @@ class Module(Definition):
     _nest = staticmethod(lambda s: {'def': Function, 'class': Class}[s])
     module = property(lambda self: self)
     all = property(lambda self: self._all)
```
keleshev: Ha-ha, lambda assignments are not allowed, but property and staticmethod assignments are fine, amirite? :smile:

pep257.py
```diff
@@ -100,7 +123,9 @@ class Module(Definition):
...
-    __str__ = lambda self: 'at module level'
+
+    def __str__(self):
+        return 'at module level'
```
keleshev: Thanks, PEP 8, this is now much more readable! :smile_cat:

sigmavirus24: It's been in the document for years now, it just wasn't checked by the tool.

pep257.py
sigmavirus24: I'm happy to change those, but the goal was to fix the test failures, not make everything more conventional.

pep257.py
sigmavirus24: @jirikuncar I'm pretty sure that'll trigger a line to long error which I think @Nurdok will still object to when formatted properly. Regardless, let me see how that looks after accounting for the default maximum line length.

sigmavirus24: Handled in dc77669

Nurdok:
Thanks @sigmavirus24 !

Issue #96, opened 2015-01-20
Allowing convenient selection of checked error codes


Nurdok:
# 0 Introduction

In this issue, I'm going to list a number of related features that I feel deserve a clear solution that can't be achieved by solving each one independently. I also suggest a possible solution. **I highly encourage anyone to criticize the features and the solution*. Of course everyone is also encouraged to suggest alternative solutions, use cases or related features. In short - show me where I'm wrong.

# 1 Features

1. Allow the user to pick a "set" of error codes to comply with a certain
   standard. For example, choosing between Google style, numpy style and PEP257
   (which should be the default behavior).
2. Allow the user to manually set which error codes are ignored when pep257 is
   run (blacklist)
3. Allow the user to manually set which error codes are considered when pep257
   is run (whitelist)
4. Allow the user to set a docsring convention (as in (1)) *and* set which
   error codes are ignored or selected *with respect to the chosen convention*.
5. Allow the user to specify an ignored error code list or convention in a
   configuration file and also *override it completely* with CLI flags.
6. Allow the user to specify an ignored error code list or conevention in a
   configuration file and also *add to it* with CLI flags.

# 2 Status Quo

Right now, the only implemented feature is (2):

    pep257 --ignore=D203,D400,D102

# 3 Proposed Solution

## 3.1 Flags

    --select=D100,D202
    --ignore=D100,D202
    --add-ignore=D100,D202
    --add-select=D100,D202
    --convention=numpy

## 3.2 Logic

1. `--select`, `--ignore` and `--convention` are mutually exclusive. It's
   illegal to specify more than one of them. However, it is possible to specify
   one such flag in the config file and another one (or even the same one) as
   a command line flag, in which case the flag specified in the config file
   will be *completely ignored*.
2. Not specifying either one of the above flags will be treated as if
   `--convention=pep257` was specified.
3. The value of the above flags (or the value of the default convention)
   determines a list of error codes that should be checked.
4. `--add-ignore` and `--add-select` can only be specified as command line
   flags. Instead of fully determining the list of errors that should be
   checked, they *alter* the list defined by the previously mentioned flags.
5. `--add-ignore` and `--add-select` are not mutually exclusive. However, 
    specifying the same error code in both is considered illegal.
6. Some error codes are mutually exclusive. After determining the final list
   of error codes that should be checked, it is considered illegal if two or
   more mutually exclusive flags in it.

# 4 Use Cases

For the sake of the following use cases, assume that there exists two mutually
[...]

PR #110, opened 2015-02-20
DRY in error definitions

Nurdok:
This PR does several things:

* Instead of each check returning an arbitrary error, all errors are defined beforehand.
* All defined errors are also put in a registry that hold them in groups (D1xx, D2xx, etc.)
* Instead of explicitly putting the error description in both the docs and the code, the error code table is generated from the error registry in the script.

The reasons are:

1. DRY (Don't Repeat Yourself). Having more than one location where errors are defined is... error prone :) And already I found that the error table from the docs had somewhat different error descriptions than the ones reported by the script.
2. Having a list of all the possible error code, instead of defining them only when an error is found by a check, will help to implement #96, by allowing to easily compile a list of selected errors whether the user specified them by a whitelist or a blacklist.


sigmavirus24:
So you actually do repeat yourself still a bit with stuff like:

```py
D1xx.create_error(...)
# 4 more times
```

Granted there will be groups (or classes) that only have one error, so `create_error` is a great method. What about, also having, `create_errors` that takes a dictionary. So the usage would look like:

```py
D1xx.create_errors({
    'D100': ('Missing docstring in public module', ),
    'D101': ('...', ),
    # ...
})
D2xx.create_errors({ ... })
```

And `create_errors` would look like:

```py
def create_errors(self, error_dictionary):
    errors = []
    for error_code in error_dictionary:
        errors.append(self.create_error(error_code, *error_dictionary[error_code]))
    return errors
```

This has the unfortunate side-effect of making it harder to change `create_error`'s signature (but you were already using positional calling for an optional parameter, so it won't make it that bad).

Nurdok:
@sigmavirus24 It did bother me a little, but the problem is that I need to assign the return value to a name, e.g.,

    D100 = D1xx.create_error('D100', 'Missing docstring in public module')

So that `D100` will be available as a return value for its corresponding check. Defining several errors in one call will not allow this, or make this cumbersome (if the call will return a tuple of errors).

sigmavirus24:
You already keep track of the errors (and keep them in an order of creation). You could use that to your advantage. You could do:

```py
D1xx.create_errors([
    ('00', ('Missing docstring in public module',)),
    ('01', ('...', )),
    # ...
])
D100, D101, D102, D103, D104 = D1xx.errors
```

Using the list of tuples will ensure the ordering is correct. Further, since `D1xx` is already created with a prefix, you can simplify how you create the errors by modifying `create_error` to do:

```py
def create_error(self, error_code, error_desc, error_context=None):
    class _Error(Error):
        code = self.prefix + error_code.lstrip(self.prefix)
        short_desc = error_desc
        context = error_context
    self.errors.append(_Error)
    return _Error
```

Using `lstrip` will allow you to not have to refactor the calls much as they are right now.

I'm not sure how I see this as being more cumbersome than the current code (in my opinion).

Nurdok:
I've given this a lot of thought and I am ambivalent about this. I see your point of the calls repeating themselves, but I find that when dealing with these sort of "declarative" code blocks, I prefer to be explicit and simple rather than efficient and complex. However minor the the differences are in this case. So I decided to go ahead with the original version - but I whole heatedly appreciate your involvement in this PR, and many others in this project. Thanks!

PR #112, opened 2015-03-12
Check that match option is applied, even on files that have been passed in

jpulec:


farmersez:
This is solved in #133.

PR #119, opened 2015-05-31
Allow docstrings given as raw unicode (ur""")

### LINKED ISSUE #116: Support ur"""docstrings"""
I have a docstring that contains both unicode characters and backslashes.

I cannot find any way to make the pep257 happy with this, though ``ur"""docstring"""`` seems suitable.

```
$ cat > test.py <<EOF
"""Some test module."""
# encoding: utf-8

def func(self):
    ur"""docstring with \backslashes and unicode, コンニチハ."""
    pass
EOF
$ pep257 test.py
test.py:4 in public function `func`:
        D301: Use r""" if any backslashes in a docstring
test.py:4 in public function `func`:
        D300: Use """triple double quotes""" (found '-quotes)
$ PAGER=cat python
>>> from test import func
>>> help(func)
Help on function func in module test:

func(self)
    docstring with \backslashes and unicode, コンニチハ.

>>>
```

jbeezley:
Fixes #116.

### DIFF
--- a/pep257.py
+++ b/pep257.py
@@ -996,11 +996,12 @@ class PEP257Checker(object):
 
         '''
         if docstring and '"""' in eval(docstring) and docstring.startswith(
-                ("'''", "r'''", "u'''")):
+                ("'''", "r'''", "u'''", "ur'''")):
             # Allow ''' quotes if docstring contains """, because otherwise """
             # quotes could not be expressed inside docstring.  Not in PEP 257.
             return
-        if docstring and not docstring.startswith(('"""', 'r"""', 'u"""')):
+        if docstring and not docstring.startswith(
+                ('"""', 'r"""', 'u"""', 'ur"""')):
             quotes = "'''" if "'''" in docstring[:4] else "'"
             return D300(quotes)
 
@@ -1014,7 +1015,8 @@ class PEP257Checker(object):
         '''
         # Just check that docstring is raw, check_triple_double_quotes
         # ensures the correct quotes.
-        if docstring and '\\' in docstring and not docstring.startswith('r'):
+        if docstring and '\\' in docstring and not docstring.startswith(
+                ('r', 'ur')):
             return D301()
 
     @check_for(Definition)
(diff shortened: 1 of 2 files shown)


Nurdok:
This PR looks good, but I'd like to see some tests for this.

jbeezley:
I originally put in some tests, but `ur` is a syntax error in python 3, so there needs to be some mechanism for adding a test file that is only executed on python 2.  Do you have any suggestions for how to do this?

Nurdok:
@jbeezley It's possible to check if the current Python is 2.x or 3.x. I think the "Pythonic" way is to check this specific syntax for `SyntaxError`, like so:

```python
try:
   eval('ur"foo"')
except SyntaxError:
    print "Python 3"
else:
    print "Python 2"
```

jbeezley:
You're right, you can catch syntax errors in an eval, but I tried various iterations of it without success.

These raise missing docstring.
```python
def raw_unicode():
    pass
raw_unicode.__doc__ = eval(r'''ur"""Test unicode \xe9 and raw \\."""''')
```
```python
def raw_unicode():
    eval(r'''ur"""Test unicode \xe9 and raw \\."""''')
```

and you can't `eval` a function definition like this.
```python
eval(r'''def raw_unicode(): ur"""Test unicode \xe9 and raw \\."""''')
```

You can `exec` a definition, but that doesn't work with your tests.  I.e.
```python
exec('def no_docstring(): pass')
```
doesn't raise an error.

Nurdok:
Instead of adding a test in `test.py`, I suggest you do it in `test_pep257.py`, where you can do something like this:

````python
def test_unicode_raw():
    if sys.version_info[0] >= 3:
        return
    with Pep257Env() as env:
        with env.open('example.py', 'wt') as example:
            example.write(textwrap.dedent("""\
                def foo():
                    ur"""docstring"""
                    pass
            """))

            # assert stuff
````

jbeezley:
@Nurdok Thanks for the hint.  I added a test.

Nurdok:
@jbeezley Merged! Thank you so much!

jquast:
Thank you @jbeezley !

PR #115, opened 2015-04-13
Add --format.

alexjurkiewicz:
Quite simple, but gives pep257 some of the flexibility of other linters (eg pylint & pep8).

I couldn't figure out how to build the docs, so I couldn't check if my docs contribution works as expected. You might need to fix it -- sorry!

PR #122, opened 2015-07-02
Feature/more options

jmullan:
Allow configuration of error templates, allow suppression of __all__ mutability error message, improve error message for wrong token kind in __all__, allow NAME tokens in __all__

In attempting to run this against a codebase, I found that I could not turn off all errors and output, and the output could not be easily parsed. I like to start with everything off, then fix one class of errors at a time.

Since 'check' was used as a global variable and inside some function scopes, I took the liberty of renaming it. I'm totally open to rewriting any of this pull request to meet your style needs. Thank you for your consideration!

### DIFF
--- a/pep257.py
+++ b/pep257.py
@@ -377,6 +377,8 @@ class Parser(object):
             elif (self.current.kind == tk.STRING or
                   self.current.value == ','):
                 all_content += self.current.value
+            elif self.current.kind == tk.NAME:
+                all_content += "'%s'" % self.current.value
             else:
                 kind = token.tok_name[self.current.kind]
                 raise AllError(
--- a/pep257.py
+++ b/pep257.py
@@ -571,6 +571,7 @@ class AllError(Exception):
             'short_desc': self._message
         }
 
+
 class ErrorRegistry(object):
     groups = []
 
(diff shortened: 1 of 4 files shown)


Nurdok:
It might take me some to get to review this, don't be discouraged by the wait :)

jmullan:
No worries!

Nurdok:
Is it possible for you to split this PR into multiple ones, each doing exactly one thing? It's very hard to read as it is. Also, regarding the error templating, there is also some work done in #115. Perhaps a better approach is to continue that work.
As a general comment, each change in behavior / API should be reflected in the docs.

Thanks!

jmullan:
Af

pep257.py
```diff
@@ -377,8 +377,15 @@ def parse_all(self):
             elif (self.current.kind == tk.STRING or
                   self.current.value == ','):
                 all_content += self.current.value
+            elif self.current.kind == tk.NAME:
+                all_content += "'%s'" % self.current.value
```
jayvdb: I believe this has been proposed before, and rejected.  I cant seem the find the issue now in any of the relevant projects.
While using symbols in __all__ may sometimes not cause a syntax error, it also doesnt work correctly.  They need to be a string.

jmullan: It was in https://github.com/GreenSteam/pep257/pull/126 , where I broke out this change in particular.

PR #125, opened 2015-08-02
Better error message for wrong token kind in __all__

jmullan:
Improves the error message when encountering an unexpected token kind in `__all__`


jmullan:
Shoot, this depends on the extended fields for AllError

jmullan:
... I took out the requirement for an extended AllErrors class.

PR #126, opened 2015-08-02
Allow NAME tokens to be put in __all__

jmullan:
This allows `__all__` to include a class directly instead of just being the string name of the class.

### DIFF
--- a/pep257.py
+++ b/pep257.py
@@ -377,6 +377,8 @@ class Parser(object):
             elif (self.current.kind == tk.STRING or
                   self.current.value == ','):
                 all_content += self.current.value
+            elif self.current.kind == tk.NAME:
+                all_content += "'%s'" % self.current.value
             else:
                 raise AllError('Could not evaluate contents of __all__. ')
             self.stream.move()


sigmavirus24:
I don't think you're supposed to put them into `__all__`. The standard library documentation demonstrates `__all__` with [strings only](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package). It seems to have been intended that way since the feature was [introduced in Python 2.1](https://docs.python.org/3/whatsnew/2.1.html).

jmullan:
I don't think you're supposed to, either, but it is legal, works, and unexpectedly blows up pep257. It seems like more of a pep8 or pylint thing to enforce.

sigmavirus24:
> but it is legal, works,

I don't think that's a good enough reason to change this behaviour.

jmullan:
I do! This is not actually a parsing error, and execution can proceed completely normally. It could be warned about, like the mutable-all, if there was some consensus that that was valuable.

keleshev:
```python
$ cat tmp.py 

def f():
    pass

__all__ = [f]
$ python
Python 2.7.8 (default, Oct 19 2014, 16:02:00) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.54)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from tmp import *
  File "<stdin>", line 1, in <module>
TypeError: attribute name must be string, not 'function'
```

jmullan:
I stand completely corrected: the offending code parses and executes just fine, and you can import individual items from a module with the erroneous definition, but you can't do `from x import *`, which is the whole point of `__all__`. (I still maintain that this is the job of pylint to handle, with E0604, but I'll close this pull request)

PR #137, opened 2015-10-02
Added D211: No blank lines allowed before class docstring

Nurdok:


Nurdok:
This is a rewrite of #91.

Issue #133, opened 2015-09-17
Config files should be read bottom-up


Nurdok:
Right now, `pep257` reads config files from the directory of the common ancestor of the checked files, and goes on up and up, reading available config files as it goes. The problem here is that the correct semantic is that the closer a config file is to the checked file, the more priority it should get.

So a config file at the top of the repo should be overridden by a config file in the same directory of the checked file.

Nurdok:
There is also another complicated issue that should be fixed regarding this. 
Consider the following directory structure:

    A
    +-- tox.ini
    +-- B
        +-- tox.ini
    
If `A/tox.ini` contains `ignore=D100` and `B/tox.ini` contains `convention=pep257`, the D100 error code **should** be checked. Meaning, it's not enough that closer-to-the-file config files should override same-flag values for more remote config files, but it should be done in a smart way such that **any** of `select`, `ignore` or `convention` override **any** of `select`, `ignore` or `convention` flags.

Nurdok:
All the while, `add-select` and `add-ignore` should be cumulative.

farmersez:
There should be a way to define "the top of the repo". My repo might be located within a "dirty" directory that contains temporary files, one of whom might be a sample `tox.ini` i'd like to keep there for some reason.

PR #138, opened 2015-10-03
#133 - Better handling of configuration from files and CLI.

farmersez:


docs/release_notes.rst
```diff
@@ -22,6 +22,9 @@ New Features
   will now be checked by default instead of D203, which required a single
   blank line before a class docstring (#137).
 
+* Now handles configuration filed correctly. The closer a configuration file
```
Nurdok: Phrase this as "Configuration files are now handled..." etc. Also filed -> files.

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
+
+    Check Configurations:
+    --------------------
+    Configurations that are related to which files and errors will be checked.
+    These are configurable in 2 ways: using the CLI, and using configuration
+    files.
+
+    Configuration files are nested within the file system, meaning that the
+    closer a configuration file is to a checked file, the more relevant it will
+    be. For instance, imagine this folder structure:
```
Nurdok: directory :)

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
-            for root, dirs, filenames in os.walk(name):
-                # Skip any dirs that do not match match_dir
-                dirs[:] = [dir for dir in dirs if match_dir(dir)]
-                for filename in filenames:
-                    if match(filename):
-                        yield os.path.join(root, filename)
+
+    CONFIG_FILE_OPTIONS = ('convention', 'select', 'ignore', 'add-select',
+                           'add-ignore', 'match', 'match-dir')
+    MUTUALLY_EXCLUSIVE_OPTIONS = ('ignore', 'select', 'convention')
```
Nurdok: These options ARE mutually exclusive, but I'm not sure if that's their defining characteristic. Perhaps it should be `BASE_ERROR_SELECTION_OPTIONS` or something similar?

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
+        self._override_by_cli = config
+
+    @check_initialized
+    def get_user_run_configuration(self):
+        """Return the run configuration for pep257."""
+        return self._run_conf
+
+    @check_initialized
+    def get_files_to_check(self):
+        """Return a generator of files and error codes to check on each file.
```
Nurdok: This _is_  a generator, it does not return a generator.

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
+        If no configuration exists (not local and not for the parend node),
+        returns and caches a default configuration.
+
+        The algorithm:
+        -------------
+        1.  If the current directory's configuration exists in
+            `self._cache` - return it.
+        2.  If a configuration file does not exist in this directory:
+        3.    If there's a parent directory:
+        4.      Cache it's configuration as this directory's and return it.
```
Nurdok: its

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
+        9.    If a parent directory exists AND the configuration file
+              allows inheritance:
+        10.     Read the parent configuration by calling this function with the
+                parent directory as `node`.
+        11.     Migrate the parent configuration with the current one and
+                cache it.
+        12. If the user has specified one of `MUTUALLY_EXCLUSIVE_OPTIONS` in
+            the CLI - return the CLI configuration with the configuration match
+            clauses
+        13. Set the `--add-select` and `--add-ignore` CLI configurations.
```
Nurdok: Replace the numbers in the algorithm with bullets - it will be much easier to maintain if it changes in the future.

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
+
+        If no configuration exists (not local and not for the parend node),
+        returns and caches a default configuration.
+
+        The algorithm:
+        -------------
+        1.  If the current directory's configuration exists in
+            `self._cache` - return it.
+        2.  If a configuration file does not exist in this directory:
+        3.    If there's a parent directory:
```
Nurdok: i.e., when it's not "/" or "C:\" ?

docs/snippets/config.rst
```diff
@@ -1,13 +1,53 @@
...
+* ``match``
+* ``match_dir``
+
+See the :ref:`cli_usage` section for more information.
+
+Inheritance
+###########
+
+By default, when finding a configuration file, ``pep257`` tries to inherit
+the parent directory's configuration and migrate them to the local ones.
```
Nurdok: The word "migrate" is used in a really weird way throughout this PR. How about "merge" instead?

src/pep257.py
farmersez: Yes. Should i add that as an example to the documentation?

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
+    @classmethod
+    def _get_checked_errors(cls, options):
+        """Extract the codes needed to be checked from `options`."""
+        checked_codes = cls._get_exclusive_error_codes(options)
+        if checked_codes is None:
+            checked_codes = cls.DEFAULT_CONVENTION
+
+        cls._set_add_options(checked_codes, options)
+
+        return checked_codes - set('')
```
Nurdok: ?

src/pep257.py
```diff
@@ -679,73 +678,496 @@ def __getattr__(self, item):
...
+            if value is not None:
+                value = set(value.split(','))
+                setattr(options, opt, value)
+
+        for opt in mandatory_set_options:
+            value = getattr(options, opt)
+            if value is None:
+                value = ''
+
+            if not isinstance(value, set):
```
Nurdok: Instead of checking specifically for `set`, check if `isinstance(value, collections.Set)`, so that it checks for the abstract interface.

src/pep257.py
```diff
@@ -854,68 +1233,40 @@ def filter(self, record):
...
-    checked_codes = get_checked_error_codes(options)
-    errors = check(collected, select=checked_codes)
+    Error.explain = run_conf.explain
+    Error.source = run_conf.source
+
+    errors = []
+    try:
+        for filename, checked_codes in conf.get_files_to_check():
+            errors.extend(check((filename,), select=checked_codes))
+    except IllegalConfiguration:
```
Nurdok: In what circumstance would this happen?

src/tests/test_pep257.py
```diff
@@ -124,33 +167,23 @@ def foo():
...
             """))
 
-        env.write_config(ignore='D100', verbose=True)
-        out, err, code = env.invoke_pep257()
+        env.write_config(ignore='D100')
+        _, err, code = env.invoke_pep257()
         assert code == 1
         assert 'D100' not in err
         assert 'D103' in err
-        assert 'example.py' in out
```
Nurdok: There should be a new test that checks the verbose flag on CLI.

src/tests/test_pep257.py
```diff
@@ -367,3 +400,440 @@ def foo():
         assert 'D100' in err
         assert 'D211' in err
         assert 'D203' not in err
+
+
+def test_config_file_inheritance():
+    """Test configuration files inheritance.
+
+    The test creates 3 configuration files:
```
Nurdok: 2?

src/tests/test_pep257.py
```diff
@@ -367,3 +400,440 @@ def foo():
...
+    env_base
+    +-- tox.ini
+    |   This configuration will set `select=`.
+    +-- A
+        +-- tox.ini
+        |   This configuration will set `inherit=false`.
+        +-- test.py
+            The file will contain code that violates D100,D103.
+
+    When invoking pep257 `test.py` the first config file will not set any
```
Nurdok: This description seems to not correspond with the above diagram.

src/tests/test_pep257.py
```diff
@@ -367,3 +400,440 @@ def foo():
...
+    |   This configuration will set `select=D103` and `match-dir=foo`.
+    +-- base.py
+    |   Will violate D100.
+    +-- A
+        +-- a.py
+            Will violate D100,D103.
+
+    We shall run pep257 with `--convention=pep257`.
+    We expect `base.py` to be checked and violate `D100` and that `A/a.py` will
+    not be checked because of `match-dir=B` in the config file.
```
Nurdok: B or foo?

src/pep257.py
Nurdok: Just phrase it as "if the directory is a root / top level directory"

src/pep257.py
farmersez: Not all configuration files are read in `config.parse()`, they are read on the go.
It's possible that during generation a new configuration file will be found and be illegal.
e.g:
A
+-- tox.ini - `select=D100`
+-- a.py
+-- B
     +-- tox.ini - `select=D100' and `ignore=D100`
     +-- b.py

Calling `pep257.py A` will raise only on the second iteration of that generator.
I'll add a comment.

src/pep257.py
farmersez: If a configuration was an empty list, `''.split(',')` will return `['']` - this clears this.
I moved to a more reasonable location.

Issue #60, opened 2013-12-25
It should be possible to ignore "magic methods" like __str__


Nurdok:
There's no reason to force documentation for methods such as `__str__`, `__unicode__`, `__repr__`, etc.

Nurdok:
I was thinking about how to implement this. If possible, I would like all customization for errors to happen by ignoring and selecting errors, so naturally I would like to have a new error code for magic methods
However, there is the issue of the exceptions that probably *should* be documented. We talked about `__init__`, but there maybe some others in that group - e.g., `__call__`. 

Some questions:

1. should the new error code exclude these "special" magic methods? 
2. Is it clear which magic methods are "special" in this way?

jacebrowning:
> Is it clear which magic methods are "special" in this way?

I would say any magic method that takes a variable number of arguments should be documented. The arguments to (and behavior of) methods like `__str__`, `__add__`, `__setattr__ `, et al. are well-defined in the data model: https://docs.python.org/3/reference/datamodel.html

Nurdok:
From what I see, the only magic methods with a variable number of arguments are `__init__`, `__new__` and `__cal__`. So I'm currently leaning towards creating a new "missing docstring in magic method" error that excludes just these three methods. Any objections?

Nurdok:
See a summary of magic methods here: http://www.rafekettler.com/magicmethods.html#appendix1

jacebrowning:
:+1:

jab:
I'm running the latest release (1.0), which it looks like this landed in, but I'm finding I have to explicitly pass `--add-ignore=D105` to suppress complaints about missing docstrings in my `__str__` methods. It looks like the intention of this issue was to suppress these by default. Did I miss something or misinterpret?

Nurdok:
@jab The new error is suppressed by default. It exists so that you _could_ suppress it, if you want to.

jab:
@Nurdok Did you mean *not* suppressed by default?

Nurdok:
@jab yes, edited.

jab:
Thanks for clarifying.

PR #139, opened 2015-10-09
Added D105: missing docstring in magic method

Nurdok:
As per #60.


sigmavirus24:
This is fantastic. Thanks @Nurdok

PR #140, opened 2015-10-09
Support '.pep257rc' as a configuration filename

jacebrowning:
PEP8 and PyLint use `.pep8rc` and `.pylintrc`, respectively, so I thought it would be good for PEP257 to support the “rc” convention as well.

### DIFF
--- a/docs/snippets/config.rst
+++ b/docs/snippets/config.rst
@@ -1,6 +1,6 @@
 ``pep257`` supports `ini`-like configuration files. In order for ``pep257`` to
-use it, it must be named ``setup.cfg``, ``tox.ini`` or ``.pep257`` and have
-a ``[pep257]`` section.
+use it, it must be named ``setup.cfg``, ``tox.ini``, ``.pep257``, or
+``.pep257rc`` and have a ``[pep257]`` section.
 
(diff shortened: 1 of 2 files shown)


jacebrowning:
The same result could be achieved if https://github.com/GreenSteam/pep257/issues/117 were implemented.

Nurdok:
pep8 doesn't seem to support `pep8rc`. In their documentation (http://pep8.readthedocs.org/en/latest/intro.html#configuration) they only state support in `.pep8`, `tox.ini` and `setup.cfg`, which is the same as pep257 (while obviously supporting `.pep257` instead of `.pep8`). Pylint supports _only_ `pylintrc`, which makes me think that pep257 is more similar to pep8 in this regard.

For these reasons, I'm inclined to not merge this PR, but feel free to try and convince me otherwise :)

jacebrowning:
I also just noticed that naming the file `.pep8` is deprecated. My main argument for using `.<tool>rc` verses `.tool` is that the former is clearly user configuration while the latter could possibly be a temporary file/directory used by that tool.

Nurdok:
Okay, I don't see too much harm in this. Merged.

jacebrowning:
Thanks!

PR #147, opened 2015-10-12
Run tox in parallel over different Python interpreters, in Travis-CI

Nurdok:


.travis.yml
```diff
@@ -6,6 +6,22 @@
 sudo: false
 language: python
 install: pip install tox
-script: tox
-python:
-  - 3.5
+script: tox -e $TOX_ENV
+matrix:
+  include:
+    - python: 2.6
```
sigmavirus24: If you make these `TOXENV` you only need do `tox` as the script because `tox` will pick it up for you.

Nurdok: Great, thanks!

PR #150, opened 2015-10-18
Remove `collect` from `__all__`

jayvdb:
The collect function was removed in a0dcb0c2,
however it was not removed from `__all__`.

### DIFF
--- a/src/pep257.py
+++ b/src/pep257.py
@@ -61,7 +61,7 @@ except AttributeError:
 
 
 __version__ = '0.7.1-alpha'
-__all__ = ('check', 'collect')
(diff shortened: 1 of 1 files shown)


src/pep257.py
```diff
@@ -61,7 +61,7 @@ def next(obj, default=nothing):
 
 
 __version__ = '0.7.1-alpha'
-__all__ = ('check', 'collect')
+__all__ = ('check')
```
jayvdb: bah, this is incorrect.  should be a tuple, as shown in https://github.com/GreenSteam/pep257/pull/155 by @lordmauve

PR #155, opened 2015-10-28
Improved imperative mood heuristic

### LINKED ISSUE #68: D401 implementation is very naive
I have the following docstring.

```
    @property
    def file_path(self):
        '''Alias for consistency with the rest of pocketlint.'''
```

The error is 

```
     142: D401: First line should be imperative: 'Alia', not 'Alias'
```

While the intent is noble, I think that the current implementation is not helpful.

Is there anyone using this check in production code?

If a check could not be done, I prefer to not try to automate it as it is not fun to maintain a list of ignored values..

Thanks!

lordmauve:
Add an improved imperative mood heuristic based on word lists, matched using a Porter stemmer (closes #68).

As discussed in issue #68, this implementation is expected to significantly reduce the false positive rate, by limiting the scope of words we are prepared to make an pronouncement about. While there are a vast number of English verbs we now disregard, a Pareto principle applies - 20% of verbs cover 80% of docstrings - meaning that a small wordlist is actually very effective. Additionally, of the verbs that are covered, we can now identify and warn about other conjugations of the verbs - for example, 'returned' and 'returning' in addition to 'returns'. Combined a blacklist of nouns, I believe the observed false negative rate in a typical codebase will also be lower.

Of the two word lists included, one is a list of verbs in their imperative form. Verbs are matched against this with stemming - effectively giving us a way of autocorrecting conjugated verbs to their imperative form. The other list includes common words that hint at a sentence structure that is not imperative, such as particles and common technical nouns. Both lists were sourced by scraping a codebase of some 30 million lines of Python and manually classifying the most common few hundred words. Between the two wordlists, with stemming, they cover 77% of the function docstrings in our codebase.

The imperative mood check was already limited to English and may need to be radically different to work in other languages. However, an approach of language-specific wordlists plus a language-specific stemmer is potentially more general than the English-only "ends in 's'" check. Such stemmers are readily available on PyPI.

To include the word lists in an easily maintainable form, the pep257 module has been changed to a package so that the word lists can be included as text files in package data.

This was tested with tox under Python 2.7 and Python 3.4.


src/pep257/__init__.py
```diff
@@ -61,7 +68,7 @@ def next(obj, default=nothing):
 
 
 __version__ = '0.7.1-alpha'
-__all__ = ('check')
+__all__ = ('check',)
```
jayvdb: IMO, this should be a separate pull request, fixing https://github.com/GreenSteam/pep257/pull/150

src/pep257/stemmer.py
```diff
@@ -0,0 +1,429 @@
...
+
+See also http://www.tartarus.org/~martin/PorterStemmer
+
+The algorithm as described in the paper could be exactly replicated
+by adjusting the points of DEPARTURE, but this is barely necessary,
+because (a) the points of DEPARTURE are definitely improvements, and
+(b) no encoding of the Porter stemmer I have seen is anything like
+as exact as this version, even with the points of DEPARTURE!
+
+Vivake Gupta (v@nano.com)
```
jayvdb: somewhere this should say it is in the public domain, based on https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/stemmer.py#L28

src/pep257/stemmer.py
```diff
@@ -0,0 +1,429 @@
+#!/usr/bin/env python
+
+"""Porter Stemming Algorithm
```
jayvdb: blank line after this line.

src/pep257/stemmer.py
```diff
@@ -0,0 +1,429 @@
...
+by adjusting the points of DEPARTURE, but this is barely necessary,
+because (a) the points of DEPARTURE are definitely improvements, and
+(b) no encoding of the Porter stemmer I have seen is anything like
+as exact as this version, even with the points of DEPARTURE!
+
+Vivake Gupta (v@nano.com)
+
+Release 1: January 2001
+
+Further adjustments by Santiago Bruno (bananabruno@gmail.com)
```
jayvdb: I believe this is @sbruno .  I assume his changes are also in the public domain.

sbruno: Hi, I don't know what is the issue with this. I did those changes a long time ago. They are not really useful. You may revert them if you want. If the question is regarding some license issue you can say it is the same as the module.

jayvdb: Thanks.  I was only wondering about license. ;-)

src/pep257/__init__.py
sigmavirus24: Agreed

sigmavirus24:
So I think I agree with @jayvdb that this shouldn't all be in one pull request. I think there should be three though:

1. For the `__all__` fix (fixing #150).
1. For turning it into a module instead of a single-file script.
1. For adding this improved heuristic.

lordmauve:
Agreed, will split when I next have time.

Nurdok:
Hey guys, I saw a bunch of PRs were opened. I just wanted to let you know it'll take me a week or two to get around to review them. Don't despair, and thanks for the effort!

Nurdok:
My thoughts regarding this: I don't like the thought of including the stemmer in the `pep257` repo. It's a huge block of code that is black magic (to me, anyway) and thus not maintainable. I am, however, willing to make `pep257` depend on a third party library for the stemmer. I'm not sure where the stemmer is taken from (Sphinx?), but the best solution from my perspective is to either include that as a dependency, or publish the stemmer as a standalone PyPI package and make `pep257` depend on that.

Any thoughts?

lordmauve:
I agree. It's a more maintainable approach if you are sold on the idea and
happy to take it further. If it's more an experiment looking for broader
feedback then it's perhaps better to skip the extra dependency.

I did some research previously and found there are a few stemmer libraries
on PyPI, many of which include support for a number of languages. I'd
suggest a key requirement is that it be pure Python; for some people and
platforms, binary dependencies are a real pain and a barrier to adoption.

lordmauve:
Gah, snowballstemmer doesn't work on py3.2 :sob:

lordmauve:
Brought this up to date in a new branch and opened #235, so closing this PR.

PR #156, opened 2015-10-30
Fix: __all__ should be a tuple

lordmauve:
As discussed in PR #155.


jayvdb:
:+1: and thanks for fixing the problem I introduced.

sigmavirus24:
Thanks for splitting this out @lordmauve

PR #162, opened 2015-11-24
Add special errors for __init__ method

JelteF:
I was having some issues with the some of the errors being to harsh on init methods. I use the docstrings for the `__init__` methods just for the arguments that it should take. The actual class info is then placed in the class docstring. 

This is why I wanted to be able to ignore some errors based on the `__init__` method. Otherwise I needed to start my `__init__` docstrings with a line with a dot followed by a newline. 

I'm not sure if this numbering and naming of errors is correct or if this is something that you even want merged.

### DIFF
--- a/src/pep257.py
+++ b/src/pep257.py
@@ -668,6 +668,9 @@ D210 = D2xx.create_error('D210', 'No whitespaces allowed surrounding '
                                  'docstring text')
 D211 = D2xx.create_error('D211', 'No blank lines allowed before class '
                                  'docstring', 'found %s')
+D212 = D2xx.create_error('D212', '1 blank line required between summary line '
+                                 'and description in __init__ method',
+                         'found %s')
 
 D3xx = ErrorRegistry.create_group('D3', 'Quotes Issues')
 D300 = D3xx.create_error('D300', 'Use """triple double quotes"""',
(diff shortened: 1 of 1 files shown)


Nurdok:
Hey @JelteF, thanks for your PR.
Unfortunately I disagree with the reasoning regarding this. Since `__init__` can accept custom parameters, every `__init__` method defines an API that should be documented. And if you're going to write a docstring, it makes no sense for it to have a different standard specifically for it.

JelteF:
I get that it is weird to have other rules for the `__init__` method, but I think it might be useful if you see my concrete use case.

All the three different section classes have the same `__init__` method, since it is inherrited: https://github.com/JelteF/PyLaTeX/blob/v1.0.0/pylatex/section.py. This way it can be generated to documentation like this: https://jeltef.github.io/PyLaTeX/current/pylatex/pylatex.section.html
I hope you agree that it would be pretty pointless to add a summary line that the `__init__` methods instantiates the class? And also that this would look ugly in the final documentation.

Because the class documentation is the only docstring that can be defined in two places like this. Split over the actual class docstring and the `__init__` method, I think it really is different enough.

PS. If you want this merged I'll fix the build so you can merge it, otherwise I'll just maintain my own fork.

Nurdok:
Right now I am still not convinced. I will, however, leave this open for a little while longer to see if anyone else wants to respond.

In any case, I do appreciate the time and effort you put into this. Thanks!

jayvdb:
@JelteF , you can achieve the same result using a flake8 extension I've built: https://github.com/jayvdb/flake8-putty (I've added an example for your use-case).  It is very new, and quite raw, and hasnt been tested with many versions of flake8/pep8.  I will attempt to fix any bugs you find promptly.

kantorii:
The following issue is still open at this point in time for this discussion.
https://github.com/PyCQA/pydocstyle/issues/189

PR #165, opened 2015-12-11
Added D403: First word of first line should be properly capitalized

Nurdok:
The new error is turned on by default.

### DIFF
--- a/src/pep257.py
+++ b/src/pep257.py
@@ -16,6 +16,7 @@ from __future__ import with_statement
 import os
 import sys
 import copy
+import string
 import logging
 import tokenize as tk
 from itertools import takewhile, dropwhile, chain
@@ -454,9 +455,9 @@ class Parser(object):
             docstring = self.parse_docstring()
             decorators = self._accumulated_decorators
             self._accumulated_decorators = []
-            log.debug("parsing nested defintions.")
+            log.debug("parsing nested definitions.")
             children = list(self.parse_definitions(class_))
-            log.debug("finished parsing nested defintions for '%s'", name)
+            log.debug("finished parsing nested definitions for '%s'", name)
             end = self.line - 1
         else:  # one-liner definition
             docstring = self.parse_docstring()
@@ -678,6 +679,11 @@ D401 = D4xx.create_error('D401', 'First line should be in imperative mood',
                          '%r, not %r')
 D402 = D4xx.create_error('D402', 'First line should not be the function\'s '
                                  '"signature"')
+D402 = D4xx.create_error('D402', 'First line should not be the function\'s '
+                                 '"signature"')
+D403 = D4xx.create_error('D403', 'First word of the first line should be '
+                                 'properly capitalized', '%r, not %r')
+
 
 
 class AttrDict(dict):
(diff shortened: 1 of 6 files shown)


src/pep257.py
```diff
@@ -128,6 +128,17 @@ def __iter__(self):
     def _publicity(self):
         return {True: 'public', False: 'private'}[self.is_public]
 
+    @property
+    def source(self):
+        """Return the source code for the definition."""
+        full_src = self._source[self._slice]
+
+        def predicate(line):
```
sigmavirus24: A better name might be nice, e.g., `is_comment_or_empty`. `predicate` is a bit too generic.

src/pep257.py
```diff
@@ -678,6 +689,10 @@ def to_rst(cls):
                          '%r, not %r')
 D402 = D4xx.create_error('D402', 'First line should not be the function\'s '
                                  '"signature"')
+D402 = D4xx.create_error('D402', 'First line should not be the function\'s '
```
sigmavirus24: Why did you duplicate 402 here?

src/pep257.py
```diff
@@ -1592,6 +1607,18 @@ def check_no_signature(self, function, docstring):  # def context
...
 
+    @check_for(Function)
+    def check_capitalized(self, function, docstring):
+        """D403: First word of the first line should be properly capitalized.
+
+        The [first line of a] docstring is a phrase ending in a period.
+
+        """
+        if docstring:
+            first_word = eval(docstring).split()[0]
```
sigmavirus24: We should use `ast.literal_eval` instead. I don't think we want people to have docstrings with arbitrary python code to execute in it.

src/pep257.py
Nurdok: Done.

src/pep257.py
Nurdok: By mistake, done.

src/pep257.py
Nurdok: In this case, we use `eval` on a `tk.STRING` token, so it seems like it can't possibly run arbitrary Python code, can it?

sigmavirus24: That's a good question. I'm not 100% certain. My familiar with the tokenizer library isn't so great.

myint: @Nurdok, why risk it? Is there some benefit to using `eval()` over the safer `ast.literal_eval()`?

sigmavirus24: So @Nurdok is correct, what you get back looks like

```py
(3, '"""Biz baz."""', (3, 3), (3, 17), '   """Biz baz."""\n')
```

So you're doing

```py
eval('"""Biz baz."""\n')  # => 'Biz baz.'
```

But it's functionally equivalent to `ast.literal_eval` in this case. I would still prefer `ast.literal_eval` to be used here, but I'm not as stringently against using `eval` as I was earlier.

We should probably add a comment regardless to reassure future reviewers as well.

Nurdok: I changed it to `ast.literal_eval`, thanks.

Nurdok:
@sigmavirus24 @myint Any more comments before I merge this?

myint:
Looks good to me!

Thanks

PR #173, opened 2016-01-29
Renaming to pydocstyle

Nurdok:


Nurdok:
Still need to:
- [ ] Add test for backward compatibility of config file section name
- [ ] Add test for backward compatibility of entry point (with deprecation warning).

sigmavirus24:
Triggering Travis for you. (This happened with pep8 too where Travis did not follow the rename)

sigmavirus24:
FWIW, I just purchased `pycqa.org` today and *just* (seconds ago) made `pydocstyle.pycqa.org` CNAME to `pep257.readthedocs.org` so you don't need to do anything about that. I'm testing out the CNAMEs right now with pydocstyle and flake8. We'll see what happens with the DNS propagation.

Nurdok:
Thanks. I also purchased pydocstyle.org/com a week or so ago. I'll set it up later today.

Issue #44, opened 2013-06-19
Summary line on second line should be allowed


artloder:
According to PEP 257, for multi-line docstrings, "the summary line may be on the same line as the opening quotes or on the next line". 

However, the current implementation of **pep257** requires that the summary line be on the same line as the opening quotes. For instance, the following tests with the summary line on the second line fail even though they appear valid by PEP 257 and are quite common docstring formats:

_check_ends_with_period_:

``` python
check = pep257.check_ends_with_period
s4 = '''"""
Should end with a period.
"""'''
assert not check(s4, None, None)
```

_check_blank_after_summary_:

``` python
pep257.check_blank_after_summary
s4 = '''"""
Blank line missing after one-line summary.

"""'''
assert not check(s4, None, None)
```

PR #174, opened 2016-02-20
Optional checks for multi-line docstring summary style.

toroettg:
Hello,

PEP 257 allows multi-line docstring summaries to start on the first or second line, as it is outlined in #44. I prefer one over the other and was looking for a way to perform consistency checks on my projects. Thus, I have added two optional error codes (D212, D213), which you might like to incorporate.

As those are more restrictive than PEP 257, both error codes should naturally be selected explicitly (and typically mutual exclusive) by the following. Otherwise, they are ignored.

```
pydocstyle --add-select=D212,D213 file_to_check.py
```

I have tried to touch as little as possible when implementing them. Please review if this is too cumbersome, or interferes with pydocstyle in a way I have not noticed.

Kind regards,
Tobias

### DIFF
--- a/docs/error_codes.rst
+++ b/docs/error_codes.rst
@@ -14,4 +14,5 @@ Not all error codes are checked for by default. The default behavior is to
 check only error codes that are part of the `PEP257
 <http://www.python.org/dev/peps/pep-0257/>`_ official convention.
 
-All of the above error codes are checked for by default except for D203.
+All of the above error codes are checked for by default except for D203,
+D212 and D213.
--- a/src/pydocstyle.py
+++ b/src/pydocstyle.py
@@ -683,6 +683,10 @@ D210 = D2xx.create_error('D210', 'No whitespaces allowed surrounding '
                                  'docstring text')
 D211 = D2xx.create_error('D211', 'No blank lines allowed before class '
                                  'docstring', 'found %s')
+D212 = D2xx.create_error('D212', 'Multi-line docstring summary should start '
+                                 'at the first line')
+D213 = D2xx.create_error('D213', 'Multi-line docstring summary should start '
+                                 'at the second line')
 
 D3xx = ErrorRegistry.create_group('D3', 'Quotes Issues')
 D300 = D3xx.create_error('D300', 'Use """triple double quotes"""',
@@ -707,7 +711,8 @@ class AttrDict(dict):
 
 
 conventions = AttrDict({
-    'pep257': set(ErrorRegistry.get_error_codes()) - set(['D203']),
+    'pep257': set(ErrorRegistry.get_error_codes()) - set(['D203', 'D212',
+                                                          'D213'])
 })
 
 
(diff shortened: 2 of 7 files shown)


Nurdok:
Thanks for submitting this! The concept is good, and I'll try to review the code as soon as I can.

src/tests/test_cases/optional.py
```diff
@@ -0,0 +1,26 @@
...
+
+from .expected import Expectation
+
+expectation = Expectation()
+expect = expectation.expect
+
+
+@expect('D212: Multi-line docstring summary does not start on the first line.')
+def multi_line_starts_second_line():
+    """
```
jayvdb: could you add some tests that use u"""...""" , r"""...""" ,etc.  I suspect that `if first not in ['"""', "'''"]:` wont catch them all.

jayvdb:
Fwiw, it was on my todo list; my main project wants to stablise our docstrings according to your D212 rule ;-)  So I am thrilled to see this patch.

When you get a chance, please squash your two commits into a single commit.  Thanks.

Just as an aside, I believe D213 equals H404.  https://github.com/openstack-dev/hacking/blob/master/hacking/checks/docstrings.py#L77

toroettg:
@javydb

You are right, I've added additional checks, but they quickly grow tedious. Another
abstraction layer like `@check_for(Content)`, to test only literals between docstring opening and closing, would come in handy to reduce redundancy.

Another enhancement for the test framework, which came to my mind, was to test against virtual files, whose content is generated dynamically at run time from definitions in test cases.

H404 seems to be equal to D212; I have somewhat harmonized this PR with their code, since I am reinventing the wheel already.

Kind regards,
Tobias

jayvdb:
> Another abstraction layer like @check_for(Content), 
> to test only literals between docstring opening and closing,
> would come in handy to reduce redundancy.

That would be very nice.  It would be a separate changeset.

docs/error_codes.rst
```diff
@@ -14,4 +14,4 @@ Not all error codes are checked for by default. The default behavior is to
 check only error codes that are part of the `PEP257
 <http://www.python.org/dev/peps/pep-0257/>`_ official convention.
 
-All of the above error codes are checked for by default except for D203.
+All of the above error codes are checked for by default except for D203, D212 and D213.
```
Nurdok: Please make sure you don't exceed 80 columns per line.

src/pydocstyle.py
```diff
@@ -683,6 +683,10 @@ def to_rst(cls):
                                  'docstring text')
 D211 = D2xx.create_error('D211', 'No blank lines allowed before class '
                                  'docstring', 'found %s')
+D212 = D2xx.create_error('D212', 'Multi-line docstring summary does not start '
+                                 'on the first line.')
+D213 = D2xx.create_error('D213', 'Multi-line docstring summary does not start '
+                                 'on a separate line.')
```
Nurdok: I prefer a positive phrasing for these - "Multi-line docstring summary should start on the first line" and "Multi-line docstring should start on a separate line".

src/tests/test_definitions.py
```diff
@@ -262,16 +262,30 @@ def test_pep257():
...
+        'multi_line_summary_start',
+    )
+
+    optional_codes = set(('D212', 'D213'))
+
+    configurations = (
+        (test_cases, set(ErrorRegistry.get_error_codes()) - optional_codes),
+        (option_test_cases, optional_codes)
+    )
+
```
Nurdok: Instead of all of this, you can just configure the test cases to always run on all the errors. It's okay if we'll get mutually exclusive errors in these and it will be less of a hassle.

Nurdok:
Make sure to also add your change to the changelog.

toroettg:
I have adapted the PR to your requested changes. With the integration as a regular definition test, I had to remove the Unicode related test cases, since they are invalid in the current PyPy3 (v2.4.0). Those may be added to #178 if so desired.

Also, the ```expect``` decorator gets confused by required arguments. Please review the change at ```a_following_valid_function``` for any side effects.

Nurdok:
@toroettg Merged! Thank you so much for your time and effort, and sorry for taking so long to merge this.

PR #176, opened 2016-02-28
Consider uppercase docstring prefixes at D300.

toroettg:
While PEP 257 suggests to use lowercase literals, Python allows uppercase literals as string prefixes.

Unlike D301 and 302, error code D300 is not reported if uppercase docstring prefixes are being used (e.g., `R'''...'''`). The current implementation of D300 also causes a false positive for uppercase literals with double quotes (e.g., `R"""Sum\\mary."""`), and detects the quotes incorrectly for a `ur'''...'''` docstring prefix. 

This proposes a more robust implementation for D300 and adds some uppercase tests for D301 and ~~D302~~, which they already pass.

EDIT: For the sake of simplicity, the D302 uppercase literal test will be included in another PR.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -4,6 +4,12 @@ Release Notes
 **pydocstyle** version numbers follow the
 `Semantic Versioning <http://semver.org/>`_ specification.
 
+x.x.x - Current Development Version
+-----------------------------------
+
+* The error code D300 is now also being reported if a docstring has
+  uppercase literals (``R`` or ``U``) as prefix (#176).
+
 1.0.0 - January 30th, 2016
 --------------------------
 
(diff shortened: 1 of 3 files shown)


sigmavirus24:
This looks good to me. Seems reasonable too.

docs/release_notes.rst
```diff
@@ -4,6 +4,12 @@ Release Notes
 **pydocstyle** version numbers follow the
 `Semantic Versioning <http://semver.org/>`_ specification.
 
+x.x.x - unknown date
```
Nurdok: Instead write "Current Development Version".

Nurdok:
Looks good. Just fix the small comment I made in the docs and I'll merge this.

toroettg:
Gladly done!

sigmavirus24:
Still looks good to me. I think @Nurdok wanted that entire line replaced with "Current development version" but `¯\_(ツ)_/¯`

Nurdok:
That is what I meant, but I'll just change this manually right after this merge.
Thanks, @toroettg!

PR #178, opened 2016-03-05
Check for D302 if at Python 3.3+ and using explicit unicode prefix.

toroettg:
Since Python 3.3, Unicode string literal prefixes have been reintroduced, but D302 is exclusively reported for Python 2.

This proposes an implementation to report D302 if and only if a Unicode prefix has been explicitly added to the docstring for Python 3. The behavior of D302 for Python 2 is not changed.

This also adds some test cases to the already existing Unicode integration test (pypy3 currently implements Python 3.2, only). To avoid code duplication, this has been implemented as a single test case with multiple checks. Let me know if you would like to see those in separate tests.

### DIFF
--- a/src/pydocstyle.py
+++ b/src/pydocstyle.py
@@ -1592,9 +1592,11 @@ class PEP257Checker(object):
 
         # Just check that docstring is unicode, check_triple_double_quotes
         # ensures the correct quotes.
-        if docstring and sys.version_info[0] <= 2:
-            if not is_ascii(docstring) and not docstring.startswith(
-                    ('u', 'ur')):
+        if docstring and not docstring.startswith('u'):
+            if sys.version_info < (3, 0) and not is_ascii(docstring):
+                return D302()
+
+            if sys.version_info >= (3, 3) and docstring.startswith('U'):
                 return D302()
 
     @check_for(Definition)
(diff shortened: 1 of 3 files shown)


src/pydocstyle.py
```diff
@@ -1592,9 +1592,11 @@ def check_unicode_docstring(self, definition, docstring):
 
         # Just check that docstring is unicode, check_triple_double_quotes
         # ensures the correct quotes.
-        if docstring and sys.version_info[0] <= 2:
-            if not is_ascii(docstring) and not docstring.startswith(
-                    ('u', 'ur')):
+        if docstring and not docstring.startswith('u'):
+            if sys.version_info < (3, 0) and not is_ascii(docstring):
```
Nurdok: Use `six.PY3` / `six.PY2` instead.

toroettg: Since this would add the first runtime dependency to the project, and thus break it's advertised self-containedness, I suggest you integrate the library as you best see fit. I will then rebase / adjust the PR accordingly. However, I think the benefit at this particular location would by marginally, as six does not offer minor version (range) distinction.

samj1912:
This PR has not had any updates in 3 years. Closing this for now. If there's any interest in resuming this, feel free to reopen.

PR #181, opened 2016-03-26
Stop converting the filename to lowercase

### LINKED ISSUE #179: path to file is lowercased
When I enter full path to file in output the path is lowecased. This leads to errors in for example atom-pydocstyle because it opens files which don't exists.
```
[fdobrovolny@minos sis]$ pydocstyle /home/fdobrovolny/DEV/sis/.../function_based.py
/home/fdobrovolny/dev/sis/.../function_based.py:12 in public function `paginate`:
        D400: First line should end with a period (not 's')
```
(...=path is shortened)
```
[fdobrovolny@minos sis]$ pydocstyle --version
1.0.0
```

sigmavirus24:
This leads to problems for tools integrating with pydocstyle, e.g., atom
linter-pydocstyle

Closes #179 

cc @BrnoPCmaniak

### DIFF
--- a/src/tests/test_integration.py
+++ b/src/tests/test_integration.py
@@ -21,7 +21,7 @@ from .. import pydocstyle
 __all__ = ()
 
 
-class TestEnv(object):
+class FakeTestEnv(object):
     """An isolated environment where pydocstyle can be run.
 
     Since running pydocstyle as a script is affected by local config files,
@@ -109,7 +109,7 @@ def install_package(request):
 
 @pytest.yield_fixture(scope="function", params=('pydocstyle', 'pep257'))
 def env(request):
-    with TestEnv(request.param) as test_env:
+    with FakeTestEnv(request.param) as test_env:
         yield test_env
 
 pytestmark = pytest.mark.usefixtures("install_package")
@@ -854,7 +854,7 @@ def test_config_file_nearest_match_re(env):
 
 
 def test_pep257_entry_point():
-    with TestEnv('pep257') as env:
+    with FakeTestEnv('pep257') as env:
         _, err, code = env.invoke()
         assert code == 0
         assert 'Deprecation Warning' in err, err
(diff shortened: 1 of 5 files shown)


Nurdok:
Notice that the tests fail for the PR - see the Travis build.
Also, this should be checked on a Windows machine - I think the `lower` call was made because there where "c:\" - "C:\" incosistencies. I'll have to check this later.

sigmavirus24:
> Notice that the tests fail for the PR - see the Travis build.

I did notice that and I was attempting to determine why before I had to leave yesterday. :)

> Also, this should be checked on a Windows machine 

Should we attempt to set up AppVeyor for pydocstyle?

For whatever it may be worth, pycodestyle doesn't munge the paths for case sensitivity and doesn't appear to have problems on Windows.

sigmavirus24:
So I updated the test case that was failing. I also updated the tests for a few things I noticed (that I'll happily pull into a separate pull request) including parametrizing the test that was failing and fixing a warning that pytest wasn't printing by default (and printing pytest warings by default).

src/tests/test_integration.py
```diff
@@ -21,7 +21,7 @@
 __all__ = ()
 
 
-class TestEnv(object):
+class FakeTestEnv(object):
```
Nurdok: Why this name change?

sigmavirus24: So I'm happy to pull d99705e into a separate pull request. I just noticed it and fixed it while working on this.

Nurdok: Oh, this is because `pytest` tries to collect this class? Then it's fine to change the name. However, I don't like "fake" because is creates a real file system environment. How about `InvocationEnv` / `StagingEnv` / `DeployEnv`, etc.?

The-Compiler: Yup - pytest tries to collect classes named `Test*`

Nurdok:
I merged this manually and changed `FakeTestEnv` to `SandboxEnv`. Thanks!

sigmavirus24:
Sorry about that. I was going to get to it tonight. :)

PR #183, opened 2016-04-23
Added D404: First word of the docstring should not be `This`

Nurdok:
This is intended to make docstrings shorter and to the point, as they should not start with "This module contains [..]" or "This class represents [..]".

It is unchecked by default as it is not part of the PEP257 standard.


sigmavirus24:
*This* pull request looks fine to me. The fact that it's off by default is helpful too.

PR #187, opened 2016-05-11
Proper parsing of imports to avoid misunderstanding __all__ imports

### LINKED ISSUE #182: Import of __all__ from another module confuses pep257
I'm doing some programmatic manipulation of `__all__` but simply to create a combined all for a package.

Thus:
```py
from .a import __all__ as all_a
from .b import __all__ as all_b

__all__ = all_a + all_b
```

Which to me should be fine since .a and .b are already evaluated by pep257 and have immutable `__all__` based tuples.

I get the error 
```
Could not evaluate contents of __all__. That means pep257 cannot decide which definitions are public. Variable __all__ should be present at most once in each file, in form `__all__ = ('a_public_function', 'APublicClass', ...)`. More info on __all__: http://stackoverflow.com/q/44834/. 
```

Interestingly if I replace `__all__` in the above example as follows:
```py
from .a import __all__ as all_a
from .b import __all__ as all_b

__all__ = ('l', 'm')
```

Then I get the same error.  So it actually seems like pep257 can't cope with me importing `__all__` from another module.

Nurdok:
Fixes #182

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -13,13 +13,16 @@ New Features
   the summary of a multi-line docstring starts at the first line,
   respectively at the second line (#174).
 
-* Added D404 - First word of the docstring should not be `This`. It is turned
+* Added D404 - First word of the docstring should not be "This". It is turned
   off by default (#183).
 
 Bug Fixes
 
 * The error code D300 is now also being reported if a docstring has
   uppercase literals (``R`` or ``U``) as prefix (#176).
+
+* Fixed a bug where an ``__all__`` error was reported when ``__all__`` was
+  imported from another module with a different name (#182, #187).
 
 1.0.0 - January 30th, 2016
 --------------------------
--- a/src/pydocstyle.py
+++ b/src/pydocstyle.py
@@ -520,7 +520,7 @@ class Parser(object):
         is_future_import = self.current.value == '__future__'
         self.stream.move()
         while (self.current.kind in (tk.DOT, tk.NAME, tk.OP) and
-                       self.current.value != 'import'):
+               self.current.value != 'import'):
             self.stream.move()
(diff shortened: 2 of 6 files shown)


jayvdb:
I think it would be useful to have a test case for `from x import __all__` without an `as` alias.

see here for a simple and reasonable example of that by @Suor:
https://github.com/Suor/funcy/blob/master/funcy/__init__.py
Not a particularly great example for this PR, as it is best case since it doesnt modify `__all__` after that import, which is where the problems start.

rules around modification of that imported `__all__` are probably best reported (if at all) by pyflakes or pycodestyle, but pydocstyle should at least have a predictable, reasonable, and stable result for that obscure scenario also.

Nurdok:
@jayvdb Added. Can you take a look at the changes?

jayvdb:
Sorry for the delay.

When `from x import __all__` is used, without an `as`, I think `AllError` needs to be raised.

A design aspect to consider is whether pydocstyle should allow code like https://raw.githubusercontent.com/Suor/funcy/master/funcy/__init__.py , as mentioned above.  i.e. should `AllError` only be raised if there are any symbols actually defined in the module which might need docstrings?  As that module doesnt have any possibly-public symbols, there is no need to add docstrings, irrespective of `__all__` or not, and therefore no need to raise `AllError`.  OTOH, it is bad style, and it meets the general requirements for `AllError`, so it is reasonable (and much easier) to raise the exception anyway.  But a decision needs to be made regarding that, and test should ensure that the decision made now isn't reversed later accidentally with an inadvertent breaking change.

Nurdok:
I agree about raising the error when importing `__all__`, but the code structure makes it a pretty big change. I'll merge this for now and work on a more structural change.

PR #188, opened 2016-06-03
Adding D405 and D406 to check return values/args

tescalada:
D405 will error if a function has a return statement but does not
define a return value in the docstring.

D406 will error if a function has arguments defined but not mentioned
in the docstring.

### DIFF
--- a/src/pydocstyle.py
+++ b/src/pydocstyle.py
@@ -1760,11 +1760,16 @@ class PEP257Checker(object):
         if not docstring:
             return
 
-        # ignore functions that already mention return
-        # FIXME: this is only checking if "return" is anywhere in
-        # the docstring, not if the return is properly documented.
-        if 'return' in docstring.lower():
-            return
+        return_styles = [
+            ':returns:',  # sphinx
+            'Returns:',  # napoleon
+            '@return:',  # twisted
+        ]
+
+        # ignore functions where any line starts with a valid return style
+        for line in docstring.split("\n"):
+            if line.strip().split()[:1] in return_styles:
+                return
 
         tree = ast.parse(function.source.strip())
 
--- a/src/pydocstyle.py
+++ b/src/pydocstyle.py
@@ -701,6 +701,10 @@ D403 = D4xx.create_error('D403', 'First word of the first line should be '
                                  'properly capitalized', '%r, not %r')
 D404 = D4xx.create_error('D404', 'First word of the docstring should not '
                                  'be `This`')
+D405 = D4xx.create_error('D405', 'Return value type should be mentioned.',
+                         '%r')
+D406 = D4xx.create_error('D406', 'Function arguments should be mentioned.',
+                         '%r')
 
 
 class AttrDict(dict):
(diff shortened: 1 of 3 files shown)


tescalada:
I wanted to check if you were interested in adding these checks before I go and update all the documentation to comply.

I can also split this in to two patches if you prefer.

Nurdok:
Thanks for the contribution but as always - it might take me a few weeks to
give it a review.

Nurdok:
Hi - I'm sorry, but my wife has given birth recently. It will take me a couple more weeks to get to this. My apologies.

src/pydocstyle.py
```diff
@@ -1696,18 +1700,93 @@ def check_starts_with_this(self, function, docstring):
...
+                pass
+
+        # ignore functions without docstrings
+        if not docstring:
+            return
+
+        # ignore functions that already mention return
+        # FIXME: this is only checking if "return" is anywhere in
+        # the docstring, not if the return is properly documented.
+        if 'return' in docstring.lower():
```
sigmavirus24: I'm not certain this is as effective. I think having return somewhere in the docstring could lead to false positives (i.e., someone having the word "return" in their string without specifying `Returns` or `:returns:`. What if we look for those at the start of a line instead of just checking for `return` in the string?

Here are the three types of return documentation I think we can largely expect:

Sphinx default:

```
:returns: ...
```

or 

```
:returns:
    ...
```

Napoleon (+Sphinx)

```
Returns:
    ...
```

Or Twisted/Pydoctor:

```
@return: ...
```

I think these are the three pre-dominant styles that we'd like to accommodate in the default syntax. (Others are looking at Numpy/Scipy styles.) We could instead split the docstring on new lines and check `.startswith((sphinx_style, napoleon_style, twisted_style))`?

tescalada:
I updated the code to handle the 3 common styles as you suggested.

myint:
It seems to fail under under Python 3.

```
$ cat foo.py
def foo(x=1):
    """Foo."""
    return 3

$ pydocstyle foo.py
  File "/private/tmp/pydocstyle/py/bin/pydocstyle", line 9, in <module>
    load_entry_point('pydocstyle==1.0.1rc1', 'console_scripts', 'pydocstyle')()
  File "/private/tmp/pydocstyle/py/lib/python3.5/site-packages/pydocstyle-1.0.1rc1-py3.5.egg/pydocstyle.py", line 1799, in main
  File "/private/tmp/pydocstyle/py/lib/python3.5/site-packages/pydocstyle-1.0.1rc1-py3.5.egg/pydocstyle.py", line 1334, in run_pydocstyle
  File "/private/tmp/pydocstyle/py/lib/python3.5/site-packages/pydocstyle-1.0.1rc1-py3.5.egg/pydocstyle.py", line 1271, in check
  File "/private/tmp/pydocstyle/py/lib/python3.5/site-packages/pydocstyle-1.0.1rc1-py3.5.egg/pydocstyle.py", line 1377, in check_source
  File "/private/tmp/pydocstyle/py/lib/python3.5/site-packages/pydocstyle-1.0.1rc1-py3.5.egg/pydocstyle.py", line 1787, in check_args
  File "/private/tmp/pydocstyle/py/lib/python3.5/site-packages/pydocstyle-1.0.1rc1-py3.5.egg/pydocstyle.py", line 1787, in <listcomp>
AttributeError: 'arg' object has no attribute 'id'
```

myint:
I just noticed that Travis CI already mentioned what I found. So never mind.

tescalada:
I haven't thought about it in a long time. It still seems like nice functionality to me, but it probably make sense to close it if no one else seems to be interested in it.

Nurdok:

Issue #171, opened 2015-12-29
Allow to filter some method names too (like files or dirs)


facundobatista:
This will allow the tool to be used in groups where have the convention of not documenting methods like __init__, or "private_*", or whatever weird name they may use.

So, being able to pass something like "--ignore-methods=foo*" or "--ignore-methods=__*" would be great.

Thanks!

akx:
:+1: for this feature.

Eric89GXL:
This would be useful, I might take a shot at implementing it since we need it for `pydocstyle` to be useful for us

Nurdok:
It might be reasonable to allow the user to specify a regex for what is considered "private" and make it default to `^_.*` (i.e., "starts with an underscore").

jayvdb:
If the equivalent of the  python 3 qual name was used, this would provide a workaround for #199.
The default regex would need to be roughly `^(.*\.)?_.*$`

PR #203, opened 2016-08-30
Allow per-function skips

### LINKED ISSUE #97: ignore error on a per line basis
pep8 supports both # noqa and # nopep8 at the end of a line to silence errors regarding that line.

It would be nice if pep257 had a similar capability to silence errors for a specific docstring, so that the entire file doesnt need to be excluded , or the same error in all files doesnt need to be excluded.

e.g. adding '# nopep257' after the docstring

### LINKED ISSUE #202: Add ignore comments
In most packages like `pyflakes`, `pep8`, and `flake8` you can do something like a `# noqa`:

```
bad_style = whatever  # noqa
```

It seems like there should be something like this for `pydocstyle`. It would allow temporary workarounds (or maybe permanent enough ones!) for things like #68, #171, #189, and any other shortcomings that may be present.

I'm happy to try adding this. I'd probably try to follow what `flake8` does since it seems pretty complete:

* `# noqa` no checks
* `# pydocstyle: noqa` no checks
* `# noqa: D200` ignore a specific error or comma-separated list of errors

Eric89GXL:
Allows per-function skips. Ready for review/merge from my end.

I added tests and updated docs in hopefully correct ways.

Closes #202.
Closes #97.


sigmavirus24:
I'm starting to think that certain things should be part of reusable library code because now pycodestyle, pydocstyle, and flake8 have just slightly different implementations of almost the same thing and that's going to lead to really weird inconsistencies between the three.

Further, from what I can tell, this will *directly* conflict with people running pydocstyle from inside Flake8 (e.g., via flake8-docstrings), which is something I'm *strongly* against.

Eric89GXL:
> Further, from what I can tell, this will directly conflict with people running pydocstyle from inside Flake8 (e.g., via flake8-docstrings), which is something I'm strongly against.

How do you mean? `# noqa` will turn off both flake8 and pydocstyle, whereas doing something like `# pydocstyle: D101` should only affect `pydocstyle`, and in theory something like `# pydocstyle: D101 flake8: E204` should work fine too.

Eric89GXL:
> I'm starting to think that certain things should be part of reusable library code because now pycodestyle, pydocstyle, and flake8 have just slightly different implementations of almost the same thing and that's going to lead to really weird inconsistencies between the three.

This might be true, but I'd prefer to do that as a separate subsequent refactoring. It will be hard to get all these libraries to settle on and adopt a new shared API, and in the meantime, something like this is badly needed for this library to be usable for many people.

Eric89GXL:
Actually it looks like `flake8` offers:

1. `# noqa` to ignore everything (and really this probably *should* cause `pydocstyle` to ignore it, too)
2. `# noqa: E501` so we can easily piggyback on that with e.g. `# noqa: E501,D101` in a compatible way. I tested this in `flake8` and it works fine.
3. `# flake8: noqa` we can again piggyback with `# flake8: noqa pydocstyle:noqa` in a compatible way. I tested this with `flake8` and it works fine.

So I don't see any conflicts in this scheme...?

docs/release_notes.rst
```diff
@@ -16,6 +16,13 @@ New Features
 * Added D404 - First word of the docstring should not be "This". It is turned
   off by default (#183).
 
+* Added the ability to ignore specific function and method doctstrings with
+  comments:
+
+    1. "# noqa" or "# pydocstyle: noqa" skips all checks.
```
sigmavirus24: @Eric89GXL This and `# pydocstyle: D102,D203` are what I'm talking about. You should just use `# noqa` below. Even so, having `# pydocstyle: noqa` in a file will skip the whole file? That's what people are going to expect from having used `# flake8: noqa`. You're going to confuse people with this.

sigmavirus24:
> This might be true, but I'd prefer to do that as a separate subsequent refactoring.

Absolutely, but that means the logic in this PR would only complicate that because it's completely turning the semantics of these comments on their head. I don't see a need to use `# pycodestyle: <codes>` as the prefix here. `pydocstyle` and `flake8` have an overlap in developers. Reusing `# noqa: <codes>` seems reasonable. It will also keep comments succinct.

docs/release_notes.rst
Eric89GXL: No it won't skip the whole file, just the line in question. I'll make it do
what flake8 does

Eric89GXL:
@sigmavirus24 Got rid of `# pydocstyle: noqa` and it should be fully compatible with `flake8` now, feel free to see if you're happy with it

jayvdb:
Do you really need to invoke `pydocstyle` directly, and add noqa support into `pydocstyle` ?

Could you use `flake8-docstrings` which gives you `# noqa` or also add [`flake8-putty`](https://github.com/jayvdb/flake8-putty) to give you `# flake8: disable=<codes>` (but is customisable to support `# noqa: <codes>`) ?

Eric89GXL:
I'll try those alternatives

Eric89GXL:
@jayvdb I tried `pip install flake8-docstrings` and then doing `flake8` instead of `pydocstyle`. It *does* respect `# noqa` and `# noqa: D103`, which helps. But at least at a first pass it has some drawbacks out of the box:

1. It doesn't respect the `pydocstyle` set of skips/includes in `setup.cfg`. So I'm not sure how to have `flake8` run on one set of files for style and another set of files for `pyflakes` and `pep8` without having to pass arguments to each (cumbersome and defeats purpose of config).

2. Installing it this way makes `flake8` by default always do `pydocstyle` checks, which is undesirable. I assume there is a way to have it "off" by default (?), but this again adds barrier to entry with every additional step necessary to configure their system.

I assume there are workarounds for this, but isn't it less clean than having `pydocstyle` respect `# noqa` and `# noqa: <codes>` when invoked on its own? After all, isn't `pydocstyle` designed to be run on its own?  The documentation makes it seem that way. With this PR, I can just tell people "run `pip install flake8 pydocstyle; flake8; pydocstyle`" and everything will work properly, and it doesn't seem so obvious with the `flake8` case. But it seems like there is sustained opposition to having this functionality here, so I'll drop it unless you're enthusiastic about it.

jayvdb:
`flake8-putty` solves 1.
You can disable `pydocstyle` by default, using `ignore=D`, but again you probably wont need this if you use flake8-putty to decide what parts of the code that `pydocstyle` should be used or ignored.

sigmavirus24:
@Eric89GXL note, that there are many teams in the PyCQA and @jayvdb is a member of one (which is not the team that maintains pydocstyle).

My opposition was merely in having divergent behaviour between two projects which you seem to have resolved.

If you have need for specific configuration from pydocstyle in flake8's configuration through `flake8-docstrings` you can create an issue on https://gitlab.com/pycqa/flake8-docstrings and we can work towards fixing those deficiencies you're finding.

Eric89GXL:
But maybe I'm missing how simple your alternative proposal is. Can you
write the docs to describe the process you mean, since they should probably
be included anyway?

Eric89GXL:
So in other words, once I have flake8-docstyle installed, I have to change
existing workflows? This seems inferior to having pydocstyle support
reasonable options out of the box without needing to install two or three
other libraries and change calls and configurations in non obvious ways.
I'm not sure why you'd want to promote this as the standard for how to
achieve this functionality.

Why such strong opposition to this? Maintenance burden?

Eric89GXL:
So this PR's documentation / API proposal was:
```
``pydocstyle`` inline commenting to skip specific checks on specific
functions or methods. The supported comments that can be added are:

1. ``"# noqa"`` skips all checks.

2. ``"# noqa: D102,D203"`` can be used to skip specific checks. Note that
   this is compatible with skips from `flake8 <http://flake8.pycqa.org/>`_,
   e.g. ``# noqa: D102,E501,D203``.
```
Since it's not being used, I guess the doc should have something like this added to show people how to use it with current `master`:
```
3. Install `flake8-docstrings`

4. Change any existing `flake8` calls on your system have to have `--ignore=D`, as installation of `flake8-docstrings` will enable these new checks by default.

5. Install `flake8-putty`.

6. Configure `flake8-putty` for each repo to triage which files get passed to `flake8` checking and `pydocstyle` checking, by <doing something>
```
I think step (4) is especially painful for devs who work on multiple repos, many of which use `flake8` (I know many such devs). It discourages actually using `pydocstyle`. It's also odd that users need to go through all this to get one type of skip functionality, when other types (like config types) are supported directly by `pydocstyle` out of the box.

Nurdok:
Hi folks - sorry for joining this party so late :)

@sigmavirus24 - it seems like the issues you mentioned have been resolved (not using `pydocstyle` in the ignore comment). As long as this issue doesn't break `flake8`, it seems fine to me. If we can come up with a nice way to share the code for the ignoring mechanism later, I'll happily refactor this.

@Eric89GXL - thanks for the PR. I'll go ahead and give you a code review now :)

docs/usage.rst
```diff
@@ -17,3 +17,9 @@ Configuration Files
 ^^^^^^^^^^^^^^^^^^^
 
 .. include:: snippets/config.rst
+
+
+Per-file configuration
```
Nurdok: How about "In-file configuration" so it'll be more clear?

src/pydocstyle.py
```diff
@@ -95,6 +95,10 @@ def leading_space(string):
 class Value(object):
 
     def __init__(self, *args):
+        if len(self._fields) != len(args):
+            raise ValueError('got %s arguments for %s fields for %s: %s'
+                             % (len(args), len(self._fields),
```
Nurdok: Please use `format` for string formatting (and not `%`) throught the code.

src/pydocstyle.py
```diff
@@ -112,7 +116,7 @@ def __repr__(self):
 class Definition(Value):
 
     _fields = ('name', '_source', 'start', 'end', 'decorators', 'docstring',
-               'children', 'parent')
+               'children', 'parent', 'skips')
```
Nurdok: Change `skips` to `skipped_error_codes` to make this more readable.

src/pydocstyle.py
```diff
@@ -462,7 +469,13 @@ def parse_definition(self, class_):
             self.leapfrog(tk.OP, value=":")
         else:
             self.consume(tk.OP)
+        skips = ''
         if self.current.kind in (tk.NEWLINE, tk.COMMENT):
+            if self.current.kind == tk.COMMENT:
+                if 'noqa: ' in self.current.value:
+                    skips = ''.join(self.current.value.split('noqa: ')[1:])
+                elif self.current.value.startswith('# noqa'):
+                    skips = 'all'
```
Nurdok: Do this in a method please (`parse_skip_comment` or similar).

src/pydocstyle.py
```diff
@@ -1400,7 +1413,10 @@ def check_source(self, source, filename):
             for check in self.checks:
                 terminate = False
                 if isinstance(definition, check._check_for):
-                    error = check(None, definition, definition.docstring)
+                    if definition.skips != 'all':
+                        error = check(None, definition, definition.docstring)
+                    else:
+                        error = None
                     errors = error if hasattr(error, '__iter__') else [error]
                     for error in errors:
```
Nurdok: You should check if `error` should be skipped _here_, instead of in each individual check function (it will save a lot of boilerplate code below).

Eric89GXL: Ahh yes that's much cleaner

Eric89GXL:
Thanks fro the comments @Nurdok, comments addressed. I'm happy to squash to a single commit so I avoid touching a bunch of now-unchanged lines if it helps.

docs/usage.rst
```diff
@@ -17,3 +17,9 @@ Configuration Files
 ^^^^^^^^^^^^^^^^^^^
 
 .. include:: snippets/config.rst
+
+
+In-file configuration
+^^^^^^^^^^^^^^^^^^^^^
+
+.. include:: snippets/per_file.rst
```
Nurdok: The snippet name should also be changed.

src/pydocstyle.py
```diff
@@ -140,13 +144,17 @@ def is_empty_or_comment(line):
         return ''.join(reversed(list(filtered_src)))
 
     def __str__(self):
-        return 'in %s %s `%s`' % (self._publicity, self._human, self.name)
+        out = 'in %s %s `%s`' % (self._publicity, self._human, self.name)
```
Nurdok: Please change this to `format` as well while you're at it :)

Nurdok:
@Eric89GXL Just a few more comments :) I don't really mind about squashing either way.

src/pydocstyle.py
Eric89GXL: Should I just replace all instances, or might that create unnecessary merge
conflicts with other open PRs?

Nurdok: I'll only hold you to lines you changed, but if you feel like going over the code, I'd be grateful.

Eric89GXL:
@Nurdok comments addressed. The only `%` string operations left in the file are to `logging` calls, which AFAIK have to stay as `%s`-style.

Nurdok:
Merged!
@Eric89GXL - thank you so much for putting in the time and effort and especially in the face of resistence.

Eric89GXL:
No problem, thanks for the helpful discussions. 

@Nurdok the next one that would make my life easier is #171. Any thoughts over there?

PR #204, opened 2016-09-06
Allow skipping decorated functions

Eric89GXL:
@Nurdok I think you talked about adding an option like this, see if it's what you have in mind. A user can now do e.g.:

```
--ignore-decorators=wraps
```
Or in my case, I will add this to my config:
```
ignore-decorators = copy_.*_doc_to_.*
```

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -9,7 +9,6 @@ from re import compile as re
 
 from . import violations
 from .config import IllegalConfiguration
-# TODO: handle
 from .parser import (Package, Module, Class, NestedClass, Definition, AllError,
                      Method, Function, NestedFunction, Parser, StringIO)
 from .utils import log, is_blank
@@ -50,10 +49,11 @@ class PEP257Checker(object):
             for this_check in self.checks:
                 terminate = False
                 if isinstance(definition, this_check._check_for):
-                    if definition.skipped_error_codes != 'all' and \
-                            not any(ignore_decorators is not None and
-                                    len(ignore_decorators.findall(dec.name))
-                                    for dec in definition.decorators):
+                    skipping_all = (definition.skipped_error_codes == 'all')
+                    decorator_skip = ignore_decorators is not None and any(
+                        len(ignore_decorators.findall(dec.name)) > 0
+                        for dec in definition.decorators)
+                    if not skipping_all and not decorator_skip:
                         error = this_check(None, definition,
                                            definition.docstring)
                     else:
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -10,7 +10,8 @@ from re import compile as re
 from . import violations
 from .config import IllegalConfiguration
 # TODO: handle
-from .parser import *
+from .parser import (Package, Module, Class, NestedClass, Definition, AllError,
+                     Method, Function, NestedFunction, Parser, StringIO)
 from .utils import log, is_blank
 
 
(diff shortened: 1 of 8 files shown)


Eric89GXL:
If this seems to work, I'm happy to tackle #171 next, since I'll want that, too.

Nurdok:
Please merge from default, this PR is out of date with the recent refactoring.

Eric89GXL:
@Nurdok merge wasn't possible because of how bad the split was, but I started fresh and manually copied the changes into the necessary places. Tests pass locallay and Travis is happy.

src/pydocstyle/checker.py
```diff
@@ -43,36 +44,40 @@ class PEP257Checker(object):
...
+            for this_check in self.checks:
                 terminate = False
-                if isinstance(definition, check._check_for):
-                    if definition.skipped_error_codes != 'all':
-                        error = check(None, definition, definition.docstring)
+                if isinstance(definition, this_check._check_for):
+                    if definition.skipped_error_codes != 'all' and \
+                            not any(ignore_decorators is not None and
+                                    len(ignore_decorators.findall(dec.name))
+                                    for dec in definition.decorators):
```
Nurdok: This `if` is too long. Break it up by assigning to local variables, something like:
```.py
checking_for_errors = ..
skipping_decorator = ..
if checking_for_errors and not skipping_decorator:
    # run the check
```
Think about better names, though :)

src/pydocstyle/config.py
```diff
@@ -139,31 +141,42 @@ def _get_matches(config):
             match_dir_func = re(config.match_dir + '$').match
             return match_func, match_dir_func
 
+        def _get_ignore_dec(config):
```
Nurdok: I prefer if you use the full word `decorator`, as this is not a common shorthand (like `func`).

src/pydocstyle/config.py
```diff
@@ -298,14 +311,12 @@ def _merge_configuration(self, parent_config, child_options):
...
-            if child_options.match is not None else parent_config.match
-        match_dir = child_options.match_dir \
-            if child_options.match_dir is not None else parent_config.match_dir
-
-        return CheckConfiguration(checked_codes=error_codes,
-                                  match=match,
-                                  match_dir=match_dir)
+        kwargs = dict(checked_codes=error_codes)
+        for key in ('match', 'match_dir', 'ignore_decorators'):
+            kwargs[key] = getattr(child_options, key) \
```
Nurdok: You can do:
```.py
kwargs[key] = getattr(child_options, key) or getattr(parent_config, key)
```

src/pydocstyle/config.py
```diff
@@ -328,21 +339,17 @@ def _create_check_config(cls, options, use_dafaults=True):
...
 
         if cls._has_exclusive_option(options) or use_dafaults:
             checked_codes = cls._get_checked_errors(options)
 
-        return CheckConfiguration(checked_codes=checked_codes,
-                                  match=match, match_dir=match_dir)
+        kwargs = dict(checked_codes=checked_codes)
+        for key in ('match', 'match_dir', 'ignore_decorators'):
+            kwargs[key] = getattr(cls, 'DEFAULT_{0}_RE'.format(key.upper())) \
+                if getattr(options, key) is None and use_dafaults \
```
Nurdok: `use_dafaults` -> `use_defaults`

src/pydocstyle/config.py
```diff
@@ -518,12 +525,20 @@ def _create_option_parser(cls):
                      "matches all dirs that don't start with "
                      "a dot").format(cls.DEFAULT_MATCH_DIR_RE))
 
+        # Decorators
+        option('--ignore-decorators', metavar='<decorators>', default=None,
```
Nurdok: You also need to update the docs in both the CLI section and the release notes.

Nurdok: I wonder, should there be special handling of "qualified" decorators? e.g., `@wraps` vs `@functools.wraps`.

src/pydocstyle/config.py
Eric89GXL: I figured that was done intentionally previously for some reason. Happy to change it though.

Eric89GXL:
Thanks for the review @Nurdok. Rebased after conflicts and then addressed your comments.

Nurdok:
Merged. Thanks!

Issue #201, opened 2016-08-28
Errors should be printed to stdout


gvanrossum:
In #26 and #29 you changed errors to be printed to stderr. But actually for a linter that's quite inconvenient when you want to post-process the output. I checked flake8, pycodestyle, pyflakes and they all write their errors to stdout (and so does mypy).

Only errors that cause the program to exit immediately (e.g. command line syntax or can't open file) should go to stderr; the expected output (diagnostics about doc style) should go to stdout.

Please?

sigmavirus24:
@gvanrossum as a way to side-step this, is it possible for you to use flake8-docstrings?

gvanrossum:
Well, I side-stepped it using `2>&1`. :-)

And now that I've installed flake8-docstrings all my flake8 runs
automatically invoke it, which I don't appreciate. (I suppose there's a
config flag to disable it.) But it does write to stdout, and the errors are
parseable by Emacs command mode, so I'm still reasonably happy with it.

Still, I think as a matter of principle all linters should write to stdout,
not to stderr, since that is their principal output.

Nurdok:
This seems reasonble, but as always, will take me some time to get to. Thanks for opening the issue.

varunagrawal:
I got this once we touch Oct 1!

lqaz:
@varunagrawal Reminder to merge.

varunagrawal:
Pushed update to #210. Can someone review it please?

Nurdok:
@varunagrawal I assure you I'll review it, but you'll have to be patient as is might take me a while.

varunagrawal:
@Nurdok sounds good. :)

PR #210, opened 2016-10-07
Print errors to stdout

varunagrawal:
Fixes for #201

### DIFF
--- a/src/pydocstyle/cli.py
+++ b/src/pydocstyle/cli.py
@@ -55,7 +55,7 @@ def run_pydocstyle(use_pep257=False):
 
     count = 0
     for error in errors:
-        sys.stderr.write('%s\n' % error)
+        sys.stdout.write('%s\n' % error)
         count += 1
     if count == 0:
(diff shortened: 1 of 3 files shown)


src/pydocstyle/cli.py
```diff
@@ -53,7 +53,7 @@ def run_pydocstyle(use_pep257=False):
 
     count = 0
     for error in errors:
-        sys.stderr.write('%s\n' % error)
+        sys.stdout.write('%s\n' % error)
```
Nurdok: Now that errors are prints to `stdout`, notice that the test for `--count` in `test_integration.py` needs to be made more robust.

Nurdok: Instead of checking that the count number is somewhere in the output, check that the it's the last line of `stdout`.

varunagrawal: @Nurdok done.

Nurdok:
Merged. Thanks!

Issue #117, opened 2015-05-15
Command line argument to specify config file


jakirkham:
It would be nice to be able to select a config file via a command argument. A feature like this exists for pep8 so something similar would be nice.

Nurdok:
This seems reasonable. Pull Requests are welcome :)

milkey-mouse:
I've done this on my fork. Just give me until Oct 1 😀

lqaz:
@milkey-mouse Reminder to merge.

PR #208, opened 2016-10-02
Add CLI option for custom config file paths

milkey-mouse:
Added a section in the docs for it too. It overrides the generic-named config files in the custom file's folder.

PR #213, opened 2016-10-10
CLI option to specify config file

varunagrawal:
Fix for #117

### DIFF
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -195,7 +195,16 @@ class ConfigurationParser(object):
         if path in self._cache:
             return self._cache[path]
 
-        config_file = self._get_config_file_in_folder(path)
+        if self._options.config_location:
+            # Find the config directory
+            cli_config_dir = os.path.dirname(self._options.config_location)
+            if os.path.abspath(cli_config_dir) == os.getcwd():
+                config_file = self._options.config_location
+            else:
+                config_file = self._get_config_file_in_folder(path)
+        else:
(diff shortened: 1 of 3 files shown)


src/pydocstyle/config.py
```diff
@@ -195,7 +195,16 @@ def _get_config(self, node):
...
             return self._cache[path]
 
-        config_file = self._get_config_file_in_folder(path)
+        if self._options.config_location:
+            # Find the config directory
+            cli_config_dir = os.path.dirname(self._options.config_location)
+            if os.path.abspath(cli_config_dir) == os.getcwd():
+                config_file = self._options.config_location
+            else:
+                config_file = self._get_config_file_in_folder(path)
```
Nurdok: I don't understand this logic. Why do you only allow configuration files from the current directory?

src/pydocstyle/config.py
```diff
@@ -195,7 +195,16 @@ def _get_config(self, node):
         if path in self._cache:
             return self._cache[path]
 
-        config_file = self._get_config_file_in_folder(path)
+        if self._options.config_location:
+            # Find the config directory
+            cli_config_dir = os.path.dirname(self._options.config_location)
+            if os.path.abspath(cli_config_dir) == os.getcwd():
```
Nurdok: You need to update this function's docstring.

Nurdok: Please add tests for this feature.

Nurdok:
Actually, it looks like this is a duplicate of #208, so I'm closing this one.

Issue #168, opened 2015-12-16
Invalid syntax exception not caught by pep257 runner


jayvdb:
When invoked directly on the command line to parse a module with invalid syntax, an exception is raised.
Not a big problem when only one module is being checked, but it causes a halt if many modules are being checked.  It would be helpful for the exception to indicate which module failed, and skip it.  It is possible to determine the failing module by using `-v`, however that requires re-running the job when an exception is encountered.

Each of the following are probably separate issues that can be solved by hardening the parser, but there will always be new syntax issues, and the pep257 runner should handle them more reasonably.

pep8 test file [E90.py](https://github.com/PyCQA/pep8/blob/fed43c5/testsuite/E90.py):

```
pep8$ pep257 testsuite/E90.py 
Traceback (most recent call last):
  File "/usr/bin/pep257", line 9, in <module>
    load_entry_point('pep257==0.7.1a0', 'console_scripts', 'pep257')()
  File "/.../pep257.py", line 1642, in main
    sys.exit(run_pep257())
  File "/.../pep257.py", line 1300, in run_pep257
    errors.extend(check((filename,), select=checked_codes))
  File "/.../pep257.py", line 1241, in check
    for error in PEP257Checker().check_source(source, filename):
  File "/.../pep257.py", line 1338, in check_source
    module = parse(StringIO(source), filename)
  File "/.../pep257.py", line 276, in __call__
    return self.parse_module()
  File "/.../pep257.py", line 428, in parse_module
    children = list(self.parse_definitions(Module, all=True))
  File "/.../pep257.py", line 370, in parse_definitions
    for definition in self.parse_definitions(class_):
  File "/.../pep257.py", line 370, in parse_definitions
    for definition in self.parse_definitions(class_):
  File "/.../pep257.py", line 378, in parse_definitions
    self.stream.move()
  File "/.../pep257.py", line 241, in move
    current = next(self._generator, None)
  File "/usr/lib64/python2.7/tokenize.py", line 357, in generate_tokens
    ("<tokenize>", lnum, pos, line))
  File "<tokenize>", line 9
    except:
    ^
IndentationError: unindent does not match any outer indentation level
```

Another one:

https://bitbucket.org/blais/snakefood/src/fa88cf9c38275362f13b208d1a7d88c4d4ab9313/test/data/simple/invalid.py

```
snakefood$ pep257 test/data/simple/invalid.py 
Traceback (most recent call last):
[...]

PR #214, opened 2016-10-11
Ignore SyntaxError while parsing

c-w:
This is a partial fix for #168. With this change, the parser no longer crashes on test file [E90.py](https://github.com/PyCQA/pep8/blob/fed43c5/testsuite/E90.py).

Note that this is a pretty minimal fix. A potential future refactor to increase the robustness of the parser could be to use the null-object pattern and return a NullToken from `TokenStream.move` instead of `None`. In this way, any code that consumes `TokenStream.current` doesn't have to null-check the property value explicitly which should make the other issues mentioned in #168 go away.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -11,6 +11,10 @@ Major Updates
 
 * Support for Python 2.6 has been dropped (#206, #217).
 
+Bug Fixes
+
+* Made parser more robust to bad source files (#168, #214)
+
 1.1.1 - October 4th, 2016
 -------------------------
 
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -209,13 +209,21 @@ class TokenStream(object):
         self._generator = tk.generate_tokens(filelike.readline)
         self.current = Token(*next(self._generator, None))
         self.line = self.current.start[0]
+        self.log = logging.getLogger()
 
     def move(self):
         previous = self.current
-        current = next(self._generator, None)
+        current = self._next_from_generator()
         self.current = None if current is None else Token(*current)
         self.line = self.current.start[0] if self.current else self.line
         return previous
+
+    def _next_from_generator(self):
+        try:
+            return next(self._generator, None)
+        except (SyntaxError, tk.TokenError):
+            self.log.warning('error generating tokens', exc_info=True)
+            return None
 
     def __iter__(self):
         while True:
--- a/src/tests/test_definitions.py
+++ b/src/tests/test_definitions.py
@@ -223,7 +223,7 @@ def test_import_parser():
             source_future_import_invalid7,
             source_future_import_invalid8,
             ), 1):
-        module = parse(StringIO(source_ucl), 'file_invalid{}.py'.format(i))
+        module = parse(StringIO(source_ucli), 'file_invalid{}.py'.format(i))
 
         assert Module('file_invalid{}.py'.format(i), _, 1,
                       _, _, None, _, _,
--- a/src/tests/test_definitions.py
+++ b/src/tests/test_definitions.py
@@ -126,6 +126,14 @@ source_future_import_invalid8 = """
 from __future__ import (, )
 """
 
+source_invalid_syntax = """
+while True:
+\ttry:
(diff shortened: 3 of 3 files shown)


sigmavirus24:
Could you add tests for this, please?

c-w:
@sigmavirus24: Sure thing. 9615c72 adds a test for the fix in dc34871. This also required making a fix in the test harness: 839b528.

src/pydocstyle/parser.py
```diff
@@ -210,14 +210,22 @@ def __init__(self, filelike):
...
+        current = self._next_from_generator()
         self.current = None if current is None else Token(*current)
         self.line = self.current.start[0] if self.current else self.line
         return previous
 
+    def _next_from_generator(self):
+        try:
+            return next(self._generator, None)
+        except SyntaxError as syntax_error:
+            self.log.warning('error generating tokens: {0}', syntax_error)
```
Nurdok: Instead of capturing the exception and passing it, you can do:
```.py
self.log.warning('error generating tokens', exc_info=True)
```

src/tests/test_definitions.py
```diff
@@ -126,6 +126,36 @@ def nested_3(self):
...
+
+# Issue #119
+# Do not crash with Python3 if the line endswith '\r\r\n'
+EMPTY_SET = set()
+SET_TYPE = type(EMPTY_SET)
+
+toto = 0 + 0
+#:
+"""  # noqa
+
```
Nurdok: Why do you need this huge snippet? You can create a syntax error with a single character (like "[").

src/pydocstyle/parser.py
c-w: Good shout.

src/tests/test_definitions.py
c-w: This snippet is the test file that was reported to cause issues in #168, but yes, we can minimize the test example. Turns out that the single invalid token (e.g. "[" like you mentioned) leads to a different exception: tk.TokenError so I added that as a test case too.

c-w:
@Nurdok: Addressed comments and resolved merge conflicts.

src/tests/test_definitions.py
```diff
@@ -126,6 +126,16 @@ def nested_3(self):
 from __future__ import (, )
 """
 
+source_invalid_syntax = """
+while True:
+\ttry:
+    pass
+"""
+
+source_token_error = """
```
Nurdok: Why do you need this to be a multi-line string? Why not just `source_token_error = "["`?

src/pydocstyle/parser.py
```diff
@@ -209,14 +209,22 @@ def __init__(self, filelike):
...
-        current = next(self._generator, None)
+        current = self._next_from_generator()
         self.current = None if current is None else Token(*current)
         self.line = self.current.start[0] if self.current else self.line
         return previous
 
+    def _next_from_generator(self):
+        try:
+            return next(self._generator, None)
+        except (SyntaxError, tk.TokenError):
```
Nurdok: Please also add this bugfix to the release notes.

src/tests/test_definitions.py
c-w: Done.

src/pydocstyle/parser.py
c-w: Added a line to `docs/release_notes.rst`.

c-w:
@Nurdok: Addressed the latest set of comments, added the fix to the release notes and did some commit squashing because the pull request was getting rather long.

docs/release_notes.rst
```diff
@@ -11,6 +11,10 @@ Major Updates
 
 * Support for Python 2.6 has been dropped (#206, #217).
 
+Bug Fixes
+
+* Made parser more robust to bad source files (#168)
```
Nurdok: Include the pull request as well. i.e., this should be:

    * Made parser more robust to bad source files (#168, #214).

c-w: Thanks. Added.

Nurdok:
Merged. Thanks!

PR #220, opened 2016-11-12
Add option to override default error output format

SpotlightKid:
This brings PR #115 up-to-date with the current sources and improves the documentation for it.

This change required two changes in some indirectly related parts of the code:

* In the Sphinx docs the intersphinx extension is enabled and configured in `conf.py`, so it is possible to link to the Python standard library documentation.
* A new `exceptions` module was added with a new exception class `InvalidErrorFormat`, to avoid circular imports from `config` and `violations`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,6 +32,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.coverage',
     'sphinx.ext.viewcode',
+    'sphinx.ext.intersphinx',  # create links to other Python documentation
     'sphinxcontrib.issuetracker',  # autolinks issue numbers (like #78)
 ]
 
@@ -261,6 +262,11 @@ texinfo_documents = [
 
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
+
+# Add sphinx.ext.intersphinx mappings
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3', None),
+}
 
 # sphinxcontrib.issuetracker settings
 issuetracker = 'github'
--- a/docs/snippets/cli.rst
+++ b/docs/snippets/cli.rst
@@ -3,6 +3,8 @@
 Usage
 ^^^^^
 
+.. highlight:: none
+
 .. code::
 
     Usage: pydocstyle [options] [<file|dir>...]
@@ -19,6 +21,8 @@ Usage
                             which errors to check for (with a list of comma-
                             separated error codes or prefixes). for example:
                             --select=D101,D2
+      --format=<format>     override default error format (ignores --explain and
+                            --source)
       --ignore=<codes>      choose the basic list of checked errors by specifying
                             which errors to ignore (with a list of comma-separated
                             error codes or prefixes). for example:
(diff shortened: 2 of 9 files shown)


SpotlightKid:
I still need to add tests, so this PR is still WIP.

I'll also have a look at allowing to specify `format` in the configuration file as well.

SpotlightKid:
I have added tests for `violations.Error.__str__`, which test both the old default behaviour and setting the error output format.

**The PR is now ready for review.**

I'll leave using `format` as a configuration option for a later PR, since that would require changing the option loading strategy.

Nurdok:
Hi @SpotlightKid, thanks for the PR. It will take me a few days to get to review this. Hold tight :)

SpotlightKid:
Ok, thanks!

sigmavirus24:
So this is the very basic kind of `--format` support that will end up frustrating about as many users as it helps. py*code*style has this and it was infuriating for users who wanted the output to be a list of json objects or a JUnit XML document. Flake8 3.x improved the status quo and made formatters pluggable and is [improving things](https://gitlab.com/pycqa/flake8/merge_requests/152) so that JUnit XML formatters are possible. It also, due to backwards compatibility concerns, handles this case as well.

I don't know how serious py*doc*style wants to get about it though.

SpotlightKid:
Sorry, I have trouble understanding the reasoning behind your comment. Why would this option be frustrating? 

I updated the earlier PR, because this option addresses a specific need of mine, namely to let my editor (Geany) be able to parse pydocstyle's output. This is just isn't possible at the moment, making working with pydocstyle rather cumbersome for me at the moment.

Also, given the speed of accepting PRs that pydocstyle currently shows, I'd rather have a basic, but functional way to format the output now, than wait for castles in the air.

sigmavirus24:
> BTW, I see no problem formatting the output as JSON, the way this option works.

Please, do show me how to format the option as an array of json objects with the basic info (error code, filename, etc.).

> Why would this option be frustrating? 

People use tools like pydocstyle in environments like Jenkins, Bamboo, Travis CI, etc. and some of those need machine-friendly output formats that can't be attained like this.

> Also, given the speed of accepting PRs that pydocstyle currently shows, I'd rather have a basic, but functional way to format the output now, than wait for castles in the air.

That's rather unfair of you to say. This project is maintained entirely by volunteers on their free time. If you dislike that @Nurdok and I have lives away from GitHub, you could better incentivize us to look at your pull request via ways other than guilt.

I was also mentioning the simplicity of another option to @Nurdok so they can determine which approach they like better.

SpotlightKid:
> Please, do show me how to format the option as an array of json objects

Ok, JSON *is* diffifult, because it is picky about quotes and commas, but YAML or XML should be doable, even though, for XML, you would need to output the enclosing elements outside of pydocstyle 

```
FMT="$(echo -e '- file: {filename}\n  line: {line}\n  code: {code}\n  short_desc: {short_desc}\n')"
pydocstyle --format="$FMT" src/tests >tests.yml
```

> That's rather unfair of you to say.

I'm just stating what I observed. I didn't mean to criticise. I also have a few small open source project, which I do in my private time and sometimes things take longer than they should. It's just the way it is, no need to get defensive about it, IMO.

Nurdok:
@SpotlightKid @sigmavirus24 
So, I went ahead and read the `flake8` formatter plugin documentation. On the whole, it seems like a good solution for formatting needs. Now I'm left with two questions - should we even _have_ a formatting option or delegate it to `flake8`? If we should, do we need it to be _as_ powerful? 

I tend to answer _yes_ to both of these. `flake8` is a great tool, but `pydocstyle` is sometimes used on its own (for example, at my workplace we don't invoke `pydocstyle` with `flake8` because it doesn't support `pydocstyle`'s configuration options).

That being said, I would like to have a formatter that is as compatible with `flake8`'s as possible. I took a look at the source code, but it seems a bit problematic as `flake8`'s `Error` class is different from `pydocstyle`'s `Error` class. However, the overall design by plugins seems logical to me.

Any thoughts on this?

SpotlightKid:
I would prefer if `pydocstyle` remained fully configurable on its own as well. As I explained, my use cases include running it manually, on-demand on a single file via an editor command.

I haven't looked at neither `flake8`'s formatter or error classes yet. I can have a look but I'M not sure when I would have the time to. OTOH, I don't see why `pydocstyle` couldn't have it's own formatter implementation.

sigmavirus24:
> Now I'm left with two questions - should we even have a formatting option or delegate it to flake8? If we should, do we need it to be as powerful?

I would answer yes to both of these as well. In no way was I trying to suggest you should leave this to Flake8. Instead, I want you to steal that from Flake8 ;). We already have seen a lot of positive feedback around that improvement in 3.x whereas we used to get frequent complaints that the `--format` parameter we inherited from py*code*style was too restrictive and not useful for anything except the simplest cases (like what's in this pull request).

> I took a look at the source code, but it seems a bit problematic as flake8's Error class is different from pydocstyle's Error class.

I'm sure we can paper around that if it turns out to be a real problem. I'm more concerned with you stealing the way it works than the exact code (although I'm sure the exact code could be helpful too).

Nurdok:
I'm closing this in favor of a more `flake8`-style solution. @SpotlightKid, sorry for the lost effort, I do appreciate you putting in the time on this. Unless you want to take up the more complete solution, I think I'll have some time for it in the upcoming days.

Issue #199, opened 2016-07-26
D100 claims module is public when it is prefixed _


jayvdb:
`Module.is_public` is hardwired to `True`, which means the following occurs

```
html5lib-python/html5lib/_inputstream.py:1:1: D100 Missing docstring in public module
```

Processing modules like this raises lots of other terminology problems, such as : should a class/def without a `_` prefix in a private module be considered "public".

A simplistic approach is to simply not process modules beginning with `_` unless they are `__init__`, both in the pydocstyle `main()` and hope flake8-docstrings does the same.

Nurdok:
Thanks for reporting this.
The simplistic approach you suggested is bad, since while a private module does not HAVE to have a docstring, if it does, it MUST be legal, so that module should be processed.
`Module.is_public` should probably change to be the same as `Class` (underscore means private).

jayvdb:
Nod.
And a non-`_`-prefixed class/def in a `_`-prefixed module...should it be treated as a public object?

Nurdok:
When checking for publicity of definitions, the logic is that it is public as long as all of its parents are public AND it doesn't begin with an underscore. So "public named" classes in a private module should be considered private and do not necessitate‬ a docstring.

Eric89GXL:
> So "public named" classes in a private module should be considered private and do not necessitate‬ a docstring.

One problem with this is that it's easy to do in `__init_.py` `from ._whatever import public_function`, and thus the function could appear public when actually used by a user (as opposed to during development). I don't see an easy solution to this problem, though.

jayvdb:
Yea. that is a common and sensible/desirable pattern.

IMO '`_`'-prefixed modules should not be processed differently. Any other approach opens a can of worms.

If the prefix means docstrings are optional, the more sensible approach is to ignose those files using config options if necessary, as they will also be skipped by most documentation tools, or writing quality docstrings and checking the module as if it was public. We could have a simple config option to ignore private modules, disabled by default which is the current behaviour.

However the messages should at least not say "public module" etc, and maybe should have different error codes for finer tuning.

varunagrawal:
I'll handle this.

varunagrawal:
@jayvdb I should just add a config option to ignore all modules beginning with a '_' and have it disabled by default?

What else would you recommend I do?

jayvdb:
[...]

PR #222, opened 2016-11-27
Modules starting with a single leading underscore are considered private

Nurdok:
Solves #199 

- [x] Add docs
- [x] Update release notes

### DIFF
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -103,10 +103,13 @@ class Module(Definition):
     _fields = ('name', '_source', 'start', 'end', 'decorators', 'docstring',
                'children', 'parent', '_all', 'future_imports',
                'skipped_error_codes')
-    is_public = True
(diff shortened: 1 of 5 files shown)


src/pydocstyle/parser.py
```diff
@@ -103,11 +103,14 @@ class Module(Definition):
...
                'children', 'parent', '_all', 'future_imports',
                'skipped_error_codes')
-    is_public = True
     _nest = staticmethod(lambda s: {'def': Function, 'class': Class}[s])
     module = property(lambda self: self)
     all = property(lambda self: self._all)
 
+    @property
+    def is_public(self):
+        return not self.name.startswith('_') or self.name.startswith('__')
```
sigmavirus24: Shouldn't this be:

```py
not (self.name.startswith('_') or self.name.startswith('__'))
```

sigmavirus24: Nevermind, just noticed the test below

PR #223, opened 2016-11-27
Drop pypy3 (temporarily)

Nurdok:
PyPy3 currently only implements the equivalent of CPython 3.2, which we don't support. Pip has also dropped support for PyPy3, resulting in the following build error:

```
/home/travis/virtualenv/pypy3-2.4.0/site-packages/virtualenv_support/pip-9.0.1-py2.py3-none-any.whl/pip/_vendor/pkg_resources/__init__.py:83: UserWarning: Support for Python 3.0-3.2 has been dropped. Future versions will fail here.
pip requires Python '>=2.6,!=3.0.*,!=3.1.*,!=3.2.*' but the running Python is 3.2.5
```

Support for PyPy3 will be restored when it is supported by Pip, and when it will implement a newer equivalent version for CPython (3.3 or higher).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ docs/_*
 
 # virtualenv
 venv/
+.venvs/
 
 # generated rst
 docs/snippets/error_code_table.rst
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,5 @@ matrix:
       env: TOXENV=py35
     - python: pypy
       env: TOXENV=pypy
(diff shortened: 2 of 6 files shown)


peterjc:
I noticed this issue noted in the v2.0.0 release notes - but PyPy3.5 v5.7 is out now and supports the equivalent of CPython 3.5

This isn't available on TravisCI by default yet, but that can be worked around - see https://github.com/travis-ci/travis-ci/issues/6727

Issue #228, opened 2017-01-21
Checks for docstring line lengths


brianv0:
PEP8 recommends doc strings actually be 72 characters or less, and it'd be nice to have a way to check for docstring length.

This may be a bit controversial as this project itself would violate the 72 character recommendation, but it does adhere to the 79 character recommendation. Meanwhile, PEP8 also mentions a relaxed 100 character line length:

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.

It may be tough to reconcile these, but I a check that verifies that all doc strings are less than 79 characters would be less controversial (as this project meets those standards) as well as a potential additional 72 character length check that could be added in the future.

peterjc:
Please support the same ``--max-line-length=n`` command line API as ``pycodestyle`` if this is added.

brianv0:
This issue was actually implemented in pycodestyle instead of pydocstyle

peterjc:
Thanks @brianv0 - cross reference issue https://github.com/PyCQA/pycodestyle/issues/344 and pull request https://github.com/PyCQA/pycodestyle/pull/674

This added new command line option ``--max-doc-length=n`` and warning ``W505 doc line too long`` which is ignored by default.

PR #229, opened 2017-01-21
Add check for 79 character line length

brianv0:
This is a pull request from #228 

This is really a PEP8 conformance, and even then this isn't checking strict conformance (72 character doc strings), but I thought maybe 79 character length strings wouldn't be to controversial.

### DIFF
--- a/src/pydocstyle/violations.py
+++ b/src/pydocstyle/violations.py
@@ -191,6 +191,7 @@ D213 = D2xx.create_error('D213', 'Multi-line docstring summary should start '
 D214 = D2xx.create_error('D214', 'Section is over-indented', '{0!r}')
 D215 = D2xx.create_error('D215', 'Section underline is over-indented',
                          'in section {0!r}')
+D216 = D2xx.create_error('D216', 'Line exceeds defined maximum length')
 
 D3xx = ErrorRegistry.create_group('D3', 'Quotes Issues')
 D300 = D3xx.create_error('D300', 'Use """triple double quotes"""',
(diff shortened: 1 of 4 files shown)


src/pydocstyle/violations.py
```diff
@@ -235,5 +236,6 @@ def __getattr__(self, item):
     'pep257': all_errors - {'D203', 'D212', 'D213', 'D214', 'D215', 'D404',
                             'D405', 'D406', 'D407', 'D408', 'D409', 'D410',
                             'D411'},
-    'numpy': all_errors - {'D203', 'D212', 'D213', 'D402'}
+    'numpy': all_errors - {'D203', 'D212', 'D213', 'D402'},
+    'pep8': all_errors - {'D216'}
```
sigmavirus24: Can you add a trailing `,` here? It will reduce future visual noise created by folks adding other style names.

Nurdok:
I wonder if we really need this. Suppose a project uses `pydocstyle` but no other linter - it would be quite weird to enforce _only_ docstrings to be less than 80 characters long, wouldn't it? And linters which enforce this limitation on the entire file would also fail for long docstrings. I'm open for discussion if someone has a good reason to add this.

@brianv0 in any case, I really appreciate the time and effort you put in. Thanks!

brianv0:
I'm think happy to go ahead and withdrawal this and submit a PR to pycodestyle instead. On second thought, it's really a PEP8 issue.

parejkoj:
Our use case is enforcing the pep8 style short docstrings, but longer code lines (e.g. 100 characters). So, there is a use for it!

Nurdok:

Issue #230, opened 2017-01-22
False positives for D410 in Numpy convention


cdeil:
@farmersez - Thank you very much for adding the Numpy convention in #226!!!

Running it on some larger projects using the Numpy convention (Numpy, Scipy, Astropy), the most common message I see is
```
 D410: Missing blank line after section
```

Personally, I don't see how requiring an empty newline at the end of the docstring is useful.

But even if that is considered part of the convention, I think pydocstyle reports false positives.

For example for https://github.com/numpy/numpy/blob/master/doc/example.py I get
```
$ pydocstyle --convention=numpy example.py
example.py:1 at module level:
        D400: First line should end with a period (not 'd')
example.py:1 at module level:
        D205: 1 blank line required between summary line and description (found 0)
example.py:1 at module level:
        D404: First word of the docstring should not be `This`
example.py:38 in public function `foo`:
        D400: First line should end with a period (not 'e')
example.py:38 in public function `foo`:
        D205: 1 blank line required between summary line and description (found 0)
example.py:38 in public function `foo`:
        D202: No blank lines allowed after function docstring (found 1)
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('Parameters')
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('Returns')
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('Other Parameters')
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('Raises')
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('See Also')
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('Notes')
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('References')
example.py:38 in public function `foo`:
        D410: Missing blank line after section ('Examples')
```

E.g. the last D410 is incorrect, I think: there is a blank line after the `Examples` section.

farmersez:
Glad you like it 😺 
Apparently I misunderstood the convention as it's described in the [HOWTO](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard) document. There's supposed to be a blank line after the whole section, not between the dashed line and the content itself. I'll upload a bugfix (hopefully) this weekend.

cdeil:
> There's supposed to be a blank line after the whole section, not between the dashed line and the content itself. 

Yes.

[...]

PR #233, opened 2017-01-27
Fix numpy d410

farmersez:


src/pydocstyle/checker.py
```diff
@@ -34,6 +39,16 @@ def decorator(f):
     return decorator
 
 
+def pairwise(iterable, default_value):
```
Nurdok: Move this to `utils.py`.

src/pydocstyle/checker.py
```diff
@@ -486,8 +501,8 @@ def _check_section_underline(cls, section_name, context, indentation):
 
         if not dash_line_found:
             yield violations.D407(section_name)
-            if next_non_empty_line_offset == 0:
-                yield violations.D410(section_name)
+            if next_non_empty_line_offset > 0:
```
Nurdok: `next_non_empty_line_offset` is a confusing name. How about something like `blank_lines_after_header`?

src/pydocstyle/checker.py
```diff
@@ -498,22 +513,24 @@ def _check_section_underline(cls, section_name, context, indentation):
...
 
-            line_after_dashes = \
-                context.following_lines[next_non_empty_line_offset + 1]
-            if not is_blank(line_after_dashes):
-                yield violations.D410(section_name)
-
             if leading_space(dash_line) > indentation:
                 yield violations.D215(section_name)
 
+            if next_non_empty_line_offset + 1 < len(context.following_lines):
```
Nurdok: If I understand correctly, this means "if this is not the last section", right? Please add a comment to explain this.

src/pydocstyle/violations.py
```diff
@@ -223,6 +223,8 @@ def to_rst(cls):
                          'Expected {0!r} dashes in section {1!r}, got {2!r}')
 D410 = D4xx.create_error('D410', 'Missing blank line after section', '{0!r}')
 D411 = D4xx.create_error('D411', 'Missing blank line before section', '{0!r}')
+D412 = D4xx.create_error('D412', 'Section content should be in the line '
```
Nurdok: I suggest making the description here clearer: "No blank lines allowed between a section header and its content".

src/tests/test_cases/sections.py
```diff
@@ -75,7 +90,7 @@ def no_blank_line_after_section():
 
     Returns
     -------
-    A whole lot of values.
+    A value of some sort.
```
Nurdok: I think the "blank line after a section" should not be enforced for the last section of the docstring (as discussed in #230).

src/pydocstyle/violations.py
farmersez: 👍

src/pydocstyle/checker.py
farmersez: 👍

src/pydocstyle/checker.py
farmersez: Legit :^)

src/pydocstyle/checker.py
farmersez: 👍

farmersez: I changed some of the logic here. The method is a bit long, but I don't think there's a non-artificial way of splitting it to several ones. Tell me if you think I should split it anyways.

src/tests/test_cases/sections.py
Nurdok: @farmersez what did you end up doing?

farmersez: I split it to a different error code which is not raised by default in the `numpy` convention.

PR #235, opened 2017-02-06
Improve imperative mood check using stemmer/wordlist

lordmauve:
This is a refresh of PR #155, to bring it up to date. This addresses issue #68. See these issues for comprehensive details of what is included in this pull request, but to summarise:

> Add an improved imperative mood heuristic based on word lists, matched using a Porter stemmer.
>
> We add a list of common verbs in the imperative form. If the stemmed form of a word matches any
> stemmed form in this list, then it's probably a verb. But if the actual form we found doesn't match the
> imperative form, we detect an error.
>
> Additionally, there is a list of words commonly found in docstrings which are not verbs.

PR #155 was out of date - it preceded the rename to pydocstyle, conversion to a package, and dropping support for Python 3.2. Git merging was practically impossible so it has been abandoned.


Nurdok:
Thanks you for opening this PR! Please be patient as it sometimes take me a while to give a proper review.

Nurdok:
On the whole, this looks good. One thing that bothers me a bit is the source of the word lists. Where did you get them from? Scraping GitHub for docstrings? If we were to make an attempt to expand the list, how would we go about that?

lordmauve:
Yes, I did it by scraping, but in a very large, proprietary codebase (tens of millions of lines), and I carefully groomed the resulting list, going from most common words downwards.

As mentioned in #155, this covered about 80% of real docstrings and was well into the long tail of words that might appear.

src/pydocstyle/violations.py
```diff
@@ -202,7 +202,9 @@ def to_rst(cls):
 D400 = D4xx.create_error('D400', 'First line should end with a period',
                          'not {0!r}')
 D401 = D4xx.create_error('D401', 'First line should be in imperative mood',
-                         '{0!r}, not {1!r}')
+                         "'{0}', not '{1}'")
+D401b = D4xx.create_error('D401', 'First line should be in imperative mood; '
```
Nurdok: I don't think this will go well when generating the docs from errors. Did you try to build the docs with this branch?

src/pydocstyle/wordlists.py
```diff
@@ -0,0 +1,39 @@
...
+    """Iterate over lines of a wordlist data file.
+
+    `name` should be the name of a package data file within the data/
+    directory.
+
+    Whitespace and #-prefixed comments are stripped from each line.
+
+    """
+    text = pkgutil.get_data('pydocstyle', 'data/' + name).decode('utf8')
+    for l in text.splitlines():
```
Nurdok: Please don't use one-letter variables - `line` is good here.

src/pydocstyle/data/imperatives_blacklist.txt
```diff
@@ -0,0 +1,109 @@
...
+# - Some adverbs don't really indicate an imperative sentence, for example
+#   "importantly" or "currently".
+# - Some irregular verb forms that don't stem to the same string as the
+#   imperative does (eg. 'does')
+:param
+:return
+:return:
+:returns
+:returns:
+:rtype
```
Nurdok: Is in necessary to put the colons here? We can just check for that when checking for D401.

lordmauve: These slipped in from the automated scanning. Should probably remove them.

src/pydocstyle/violations.py
lordmauve: I did not. The problem here is that on Python 2, unicode strings have a repr like `u'foo'` which is both unsightly and causes test failures.

src/pydocstyle/wordlists.py
lordmauve: I prefer one-letter variables when they have very ephemeral scope! But I take your point, and `l` is specifically forbidden by PEP8.

src/pydocstyle/violations.py
lordmauve: I checked, this makes no difference. Only the first and second arguments here are used in the docs.

setup.py
```diff
@@ -31,6 +31,10 @@
     keywords='pydocstyle, PEP 257, pep257, PEP 8, pep8, docstrings',
     packages=('pydocstyle',),
     package_dir={'': 'src'},
+    package_data={'pydocstyle': ['data/*.txt']},
+    install_requires=[
+        'snowballstemmer==1.2.1',
```
Nurdok: It's okay to pin version in the development requirements, but the install requirements should be as loose as possible, so please remove the version number.

lordmauve:
Thanks, I have done so.

lordmauve:
Thank you!

PR #236, opened 2017-02-10
Support custom config file path via '--config' CLI option

jacebrowning:
Closes https://github.com/PyCQA/pydocstyle/issues/117

This is based on the work by @milkey-mouse (Thanks!) in https://github.com/PyCQA/pydocstyle/pull/208 and supersedes it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ docs/_*
 
 # virtualenv
 venv/
+.venv/
 .venvs/
 
 # generated rst
--- a/docs/snippets/config.rst
+++ b/docs/snippets/config.rst
@@ -12,6 +12,9 @@ file specified above *in that exact order*. If a configuration file was not
 found, it keeps looking for one up the directory tree until one is found or
 uses the default configuration.
 
+``pydocstyle`` also accepts a ``--config`` CLI option to specify the path to a
+custom config file location.
+
 .. note::
 
     For backwards compatibility purposes, **pydocstyle** supports configuration
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -208,7 +208,10 @@ class ConfigurationParser(object):
         if path in self._cache:
             return self._cache[path]
 
-        config_file = self._get_config_file_in_folder(path)
+        if self._options.config_file:
+            config_file = os.path.abspath(self._options.config_file)
+        else:
+            config_file = self._get_config_file_in_folder(path)
 
         if config_file is None:
             parent_dir, tail = os.path.split(path)
@@ -257,6 +260,9 @@ class ConfigurationParser(object):
         parser = RawConfigParser()
         options = None
         should_inherit = True
+
+        if not os.path.isfile(path):
+            self._parser.error("Config file not found: {}".format(path))
 
         if parser.read(path) and self._get_section_name(parser):
             option_list = dict([(o.dest, o.type or o.action)
@@ -509,6 +515,8 @@ class ConfigurationParser(object):
                help='print status information')
         option('--count', action='store_true', default=False,
                help='print total number of errors to stdout')
+        option('--config', metavar='<path>', dest='config_file',
+               help='specify a custom config file location')
 
         # Error check options
         option('--select', metavar='<codes>', default=None,
--- a/docs/snippets/cli.rst
+++ b/docs/snippets/cli.rst
@@ -15,6 +15,7 @@ Usage
       -d, --debug           print debug information
       -v, --verbose         print status information
       --count               print total number of errors to stdout
+      --config=<path>       specify a custom config file location
       --select=<codes>      choose the basic list of checked errors by specifying
                             which errors to check for (with a list of comma-
                             separated error codes or prefixes). for example:
--- a/src/tests/test_integration.py
+++ b/src/tests/test_integration.py
@@ -38,7 +38,7 @@ class SandboxEnv(object):
         self.tempdir = None
         self.script_name = script_name
 
-    def write_config(self, prefix='', **kwargs):
+    def write_config(self, prefix='', filename='tox.ini', **kwargs):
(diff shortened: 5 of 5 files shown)


.gitignore
```diff
@@ -43,6 +43,7 @@ docs/_*
 
 # virtualenv
 venv/
+.venv/
```
jacebrowning: This is one of the default locations for the new `pipenv` tool.

sigmavirus24: We don't use `pipenv` though.

jacebrowning: Does it hurt anything to keep this? I can remove it if need be.

Nurdok: I don't mind it.

docs/snippets/config.rst
```diff
@@ -12,6 +12,8 @@ file specified above *in that exact order*. If a configuration file was not
 found, it keeps looking for one up the directory tree until one is found or
 uses the default configuration.
 
+``pydocstyle`` also accepts a ``--config`` CLI option to specify the path to a custom config file location.
```
Nurdok: Please make sure that lines adhere to the "80 chars" limit.

src/pydocstyle/config.py
```diff
@@ -208,7 +208,10 @@ def _get_config(self, node):
         if path in self._cache:
             return self._cache[path]
 
-        config_file = self._get_config_file_in_folder(path)
+        if self._options.config_file:
```
Nurdok: After you set the config file, this function also searches for more config files in parent directories, etc., which shouldn't happen when you pass a config file via `--config`. Instead of doing this here, I think you need to refactor this function into several small ones so you can just get the functionality you want (such as merging configuration).

jacebrowning: I don't fully understand your comment. Which parts of this function do you expect to be skipped when using the `--config` option?

maglfmn: @Nurdok: If you say ``--config=my.txt`` it would look at the current directory and then traverse up to find it, no?

Nurdok: I think I understand the confusion. In my mind, the `--config` option is meant to specify a specific config file (path to a config file), not allow a new name for config files in general. It seems more useful to me. Do you object?

Nurdok: @farmersez any input?

jacebrowning: I have not used the parent directory lookup/merge feature, so I need someone else to make the call, but I could see a use case for pointing to a project-specific configuration file while also loading shared settings for all projects.

farmersez: I don't agree @Nurdok - Imagine having 2 repos, A and B. You wish to test your code in repo B to see if it complies with the standards of repo A, so you run `pydocstyle --config-file A/subdir1/subdir2/tox.ini B/`.
If the configuration loader wouldn't inherit and merge the parent config files, pydocstyle wouldn't use the same configurations for `B/` as `A/subdir1/subdir2/` does.

Perhaps a `--dont-inherit` flag should be added?

Nurdok: If we go with this option, given `--config=my.txt` - should `pydocstyle` look for `my.txt` files all the way up, ignoring e.g., `tox.ini` files, or _start_ with _this specific_ `my.text` file and look up for "normal" config files? I tend toward the latter. In any case, whatever is decided should be well documented, as this is starting to get confusing.

farmersez: I don't think this should affect the config files search logic, only the initial directory/file the search is started from. Meaning, `--config=PATH`, where `PATH` can be a standard config file path or a directory path (so you could check if your file complies with standards of files in a directory, no matter where their configuration is in the directory tree).

jacebrowning: IMO, specifying a custom directory path (rather than a specific config file) should be a different command-line option (perhaps `--config-dir`?) and is outside the scope of this change.

farmersez: @jacebrowning , I just noticed a problem - in line 232 `self._get_config` is called recursively on the parent file in order to inherit its options, but the call is irrelevant because you'd just override the file path and use the same file from the cli (line 212).

Considering my previous comment, i think that ignoring the configuration hierarchy is definitely not the way to go since one config file is only defining a part of the final configuration.  In that case, the difference between specifying a path to a directory vs a path to a config file is little. I'd support both by calling `self._get_config_file_in_folder(self._options.config_path)`.

jacebrowning: @farmersez Won't calling `self._get_config_file_in_folder(self._options.config_path)` disable the ability to specify a custom config filename? `_get_config_file_in_folder` disregards paths to specific files and instead uses `PROJECT_CONFIG_FILES`.

Nurdok: I think letting `--config` also search up will be of no use to anyone. I can see how `--config` could be used to, for example, point to a shared config file that is not in any parent directory, but just using a different project's specific directory's configuration... is at least rare, if not meaningless. I suggest the following: 

* Make `--config` take a single, specific file. 
* Using `--config` will cancel any other configuration file search, and inheritance (i.e., it will imply `inherit = false`).
* For the case where we want to "flatten" a complex inherited configuration to a single file, we can add a flag or a script that ignores actual files and just print the computed configuration.

Thoughts?

farmersez: I take back my wanna-test-my-code-on-another-project's-configuration, it does seem a little pointless 😋 
+1 @Nurdok

Nurdok: @jacebrowning so can you change this to accept a single, specific file, and not performing further searches? If you lost interest because of #237, that's ok :)

jacebrowning: Yeah, the specifics of `--config` are more complicated that I expected and #237 works well for me. Someone else can pick this up if they feel so inclined. ✌️

PR #237, opened 2017-02-15
Allow file extensions on config files

jacebrowning:
Supporting configuration files with the `.ini` extension means that many editors will support syntax highlighting out-of-the-box without needing an additional plugin or settings tweak.


Nurdok:
LGTM. Merged.

jacebrowning:
Thanks! Personally, this was my main use case for https://github.com/PyCQA/pydocstyle/pull/236, but others might have additional reasons they want a `--config=<path>` option.

PR #240, opened 2017-03-04
WIP: Use Jedi as the underlying parser

Nurdok:
WIP, description to come.


Nurdok:
This is still in the works, but I would appreciate any thoughts or comments from anyone who has a bit of time (esp. `parser.py`).

sigmavirus24:
I haven't looked at the code at all. My first question, however, is how does a Jedi AST differ from a Python stdlib AST? I ask specifically because Flake8 defaults to using a stdlib AST. It provides that parsed AST (doing it only once) to each plugin that traverses an AST. So if pydocstyle develops its own Flake8 plugin, it'll need to know how to take a stdlib AST and use that instead of a Jedi AST (or we'll have to work on converting Flake8 to using Jedi ASTs).

setup.py
```diff
@@ -31,6 +31,9 @@
     keywords='pydocstyle, PEP 257, pep257, PEP 8, pep8, docstrings',
     packages=('pydocstyle',),
     package_dir={'': 'src'},
+    install_requires=[
+        'jedi==0.9.0',
```
sigmavirus24: Please, let's not pin this. This causes unending headaches for users and downstream redistributors.

src/__init__.py
```diff
@@ -0,0 +1 @@
+"""Pydocstyle - a docstring style checker."""
```
sigmavirus24: Is `src/` meant to be a package? If yes, why? To be clear, I don't think `src/` should be a package.

src/pydocstyle/checker.py
```diff
@@ -680,6 +681,10 @@ def check(filenames, select=None, ignore=None, ignore_decorators=None):
         try:
             with tokenize_open(filename) as file:
                 source = file.read()
+                try:
+                    source = source.decode('utf-8')
+                except:
```
sigmavirus24: I think this except should be more narrowly scoped. There are a surprising number of exceptions that can happen here and they depend on platform, how the python interpreter is compiled, etc.

src/tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Test for pydocstyle."""
```
sigmavirus24: py.test does not require an `__init__.py` and often recommends you don't make your tests module importable.

src/pydocstyle/parser.py
```diff
@@ -80,9 +94,16 @@ def _publicity(self):
         return {True: 'public', False: 'private'}[self.is_public]
 
     @property
+    def source_lines(self):
+        if '_source' in self._fields:
```
sigmavirus24: Will `_source` ever be in `_fields` now? Can't you just make this property a proxy for `self.parent.source_lines` now?

src/pydocstyle/parser.py
```diff
@@ -186,12 +207,26 @@ def is_public(self):
 class Decorator(Value):
     """A decorator for function, method or class."""
 
-    _fields = 'name arguments'.split()
+    _fields = 'name'.split()
+
+
+class DunderAll(Value):
+    """A list of exported names in a module."""
+    _fields = 'names'.split()
```
sigmavirus24:
```py
>>> 'names'.split()
['names']
```

Why not simply do?

```py
_fields = ['names']
```

src/pydocstyle/parser.py
```diff
@@ -186,12 +207,26 @@ def is_public(self):
...
+
+
+class FutureImport(Value):
+    """A list of __future__ imports in a module."""
+    _fields = 'name'.split()
```
sigmavirus24: Same question.

src/pydocstyle/parser.py
```diff
@@ -250,311 +285,278 @@ def __init__(self, *args):
...
+            return ''
+        comment = token[1]
         skipped_error_codes = ''
-        if self.current.kind == tk.COMMENT:
-            if 'noqa: ' in self.current.value:
-                skipped_error_codes = ''.join(
-                     self.current.value.split('noqa: ')[1:])
-            elif self.current.value.startswith('# noqa'):
-                skipped_error_codes = 'all'
+        if 'noqa: ' in comment:
```
sigmavirus24: If I stripped the `noqa` logic into a separate library, would you use that instead?

src/pydocstyle/checker.py
duboviy: To catch any exception use the ```except Exception```, instead of ```except BaseException``` or just ```except```.
Indicate the most specific exception type in ```except``` block.
For example:
```python
# Bad
try:
    mapping[key]
except Exception:
    ...

# Better
try:
    mapping[key]
except KeyError:
    ...
```

setup.py
Nurdok: Eventually this will not be pinned. I temporarily pinned this because a new `jedi` was released while I was working on this, which broke some of my changes.

samj1912:

PR #243, opened 2017-03-27
Backport parser tests and some changes from Jedi branch

Nurdok:
Since the Jedi team are working on an API, I'm pausing work on #240. However, I have rewritten the parser tests so I am adding them to pydocstyle in this PR. There are also some changes such as raising `ParseError` is the Python file's syntax is illegal. When this is done, a new version can be released (2.0.0).

#### To Do
- ~~Add a test that runs pydocstyle on two files and make sure they are both checked even if the first has illegal syntax.~~


src/pydocstyle/cli.py
```diff
@@ -49,7 +49,8 @@ def run_pydocstyle():
 
     count = 0
     for error in errors:
-        sys.stdout.write('%s\n' % error)
+        if hasattr(error, 'code'):
```
farmersez: I'd check that the type of `error` does not inherit from `BaseException`. Ducktyping here seems... odd (imo).

src/pydocstyle/parser.py
```diff
@@ -30,7 +29,12 @@ def next(obj, default=nothing):
 
 __all__ = ('Parser', 'Definition', 'Module', 'Package', 'Function',
            'NestedFunction', 'Method', 'Class', 'NestedClass', 'AllError',
-           'StringIO')
+           'StringIO', 'ParseError')
+
+
+class ParseError(Exception):
+    def __str__(self):
+        return "cannot parse file"
```
farmersez: `AllError`'s message is capitalized, so should this.

src/tests/parser_test.py
```diff
@@ -291,3 +429,138 @@ def test_module_publicity():
...
     assert module.is_public
+
+
+def test_complex_module():
+    """Test that a complex module is parsed correctly."""
+    parser = Parser()
+    code = CodeSnippet('''\
+        """Module."""
+        __all__ = ('a', 'b'
+                   'c',)
```
farmersez: Is this on a separate line on purpose?

src/tests/parser_test.py
```diff
@@ -291,3 +429,138 @@ def test_module_publicity():
...
+        def function():
+            "Function."
+            def nested_1():
+                """Nested."""
+            if True:
+                def nested_2():
+                    pass
+        class class_(object):
```
farmersez: `class_1`? Just an underscore at the end looks fugly 😋

src/tests/test_integration.py
```diff
@@ -954,3 +964,15 @@ def foo():
     _, _, code = env.invoke()
 
     assert code == 0
+
+
+def test_syntax_error_multiple_files(env):
```
farmersez: Perhaps add a third _valid_ file and see that it's parsed correctly? (or is that TODO in the issue? Why not do it now?)

src/pydocstyle/cli.py
sigmavirus24: Why? What does checking for that prevent?

src/tests/parser_test.py
```diff
@@ -291,3 +429,138 @@ def test_module_publicity():
...
     module = parser.parse(code, "__filepath")
     assert module.is_public
+
+
+def test_complex_module():
+    """Test that a complex module is parsed correctly."""
+    parser = Parser()
+    code = CodeSnippet('''\
+        """Module."""
+        __all__ = ('a', 'b'
```
sigmavirus24: Should there be a trailing comma here?

src/tests/test_integration.py
sigmavirus24: Well if a single syntax error prevented further checking, then wouldn't it hold that we would never see the second one produce a syntax error either? I don't think a third file is needed here.

src/tests/parser_test.py
Nurdok: Yes, it's a corner case for the parser.

src/tests/test_integration.py
Nurdok: What @sigmavirus24 said :)

src/tests/parser_test.py
Nurdok: No, it's a concatenated string.

src/pydocstyle/cli.py
Nurdok: @farmersez why is `isinstance` better? I think this clearly reads as "print any error that has an error code associated with it".

src/pydocstyle/parser.py
Nurdok: Done.

src/tests/parser_test.py
Nurdok: It's just test code in a string... I don't think it matters.

src/pydocstyle/cli.py
farmersez: `errors` contain 2 object types - Exceptions (caught by pydocstyle) and errors (generated by pydocstyle). If some day `errors` would contain more than those 2 object types the `hasattr` check might not be good enough, but my proposal might not be good as well. Leave it as it is.

PR #246, opened 2017-04-16
Added matrix multiplication operator handling

### LINKED ISSUE #191: Parsing fails with Python 3.5 matrix multiply infix operator
pydocstyle fails when parsing files that use the Python 3.5 matrix multiply infix operator `@`.
# Steps to reproduce

Create file `blah.py`:

``` python
import numpy as np
def blah_bad():
    """Blah blah bad"""

    a = np.zeros(1)
    b = np.zeros(1)
    x = a @ b
```

Run pydocstyle on `blah.py`:

``` bash
pydocstyle blah.py
```

Output:

```
  File "pydocstyle.py", line 1725, in <module>
    main()
  File "pydocstyle.py", line 1715, in main
    sys.exit(run_pydocstyle(use_pep257))
  File "pydocstyle.py", line 1330, in run_pydocstyle
    errors.extend(check((filename,), select=checked_codes))
  File "pydocstyle.py", line 1267, in check
    for error in PEP257Checker().check_source(source, filename):
  File "pydocstyle.py", line 1368, in check_source
    module = parse(StringIO(source), filename)
  File "pydocstyle.py", line 275, in __call__
    return self.parse_module()
  File "pydocstyle.py", line 428, in parse_module
    children = list(self.parse_definitions(Module, all=True))
  File "pydocstyle.py", line 366, in parse_definitions
    yield self.parse_definition(class_._nest(self.current.value))
  File "pydocstyle.py", line 485, in parse_definition
    class_.__name__, name, self.current.kind,
AttributeError: 'NoneType' object has no attribute 'kind'
```
# Source of problem

The problem is in `pydocstyle.Parser#parse_definitions` where `@` is assumed to exclusively be part of a decorator in the lines:

``` python
            elif self.current.kind == tk.OP and self.current.value == '@':
                self.consume(tk.OP)
                self.parse_decorators()
```

I'm not sufficiently intimately familiar with the valid Python syntax to suggest a fix, but you could presumably solve the problem by assuming that any `@` characters encountered while parsing an expression are matrix multiply operators if they're not the first non-whitespace character of the expression.

Nurdok:
Fixes #191.

### DIFF
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -224,21 +224,26 @@ class AllError(Exception):
 
 
 class TokenStream(object):
-    NEWLINES = {tk.NEWLINE, tk.INDENT, tk.DEDENT}
+    # A logical newline is where a new expression or statement begins. When
+    # there is a physical new line, but not a logical one, for example:
+    # (x +
+    #  y)
+    # The token will be tk.NL, not tk.NEWLINE.
+    LOGICAL_NEWLINES = {tk.NEWLINE, tk.INDENT, tk.DEDENT}
 
     def __init__(self, filelike):
         self._generator = tk.generate_tokens(filelike.readline)
         self.current = Token(*next(self._generator, None))
         self.line = self.current.start[0]
         self.log = log
-        self.got_newline = True
+        self.got_logical_newline = True
 
     def move(self):
         previous = self.current
         current = self._next_from_generator()
         self.current = None if current is None else Token(*current)
         self.line = self.current.start[0] if self.current else self.line
-        self.got_newline = (previous.kind in self.NEWLINES)
+        self.got_logical_newline = (previous.kind in self.LOGICAL_NEWLINES)
         return previous
 
     def _next_from_generator(self):
@@ -275,7 +280,6 @@ class Parser(object):
 
     def parse(self, filelike, filename):
         """Parse the given file-like object and return its Module object."""
-        # TODO: fix log
         self.log = log
         self.source = filelike.readlines()
         src = ''.join(self.source)
@@ -380,12 +384,12 @@ class Parser(object):
         while self.current is not None:
             self.log.debug("parsing definition list, current token is %r (%s)",
                            self.current.kind, self.current.value)
-            self.log.debug('got_newline: %s', self.stream.got_newline)
+            self.log.debug('got_newline: %s', self.stream.got_logical_newline)
             if all and self.current.value == '__all__':
                 self.parse_all()
             elif (self.current.kind == tk.OP and
                   self.current.value == '@' and
-                  self.stream.got_newline):
+                  self.stream.got_logical_newline):
                 self.consume(tk.OP)
                 self.parse_decorators()
             elif self.current.value in ['def', 'class']:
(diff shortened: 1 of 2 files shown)


src/pydocstyle/parser.py
```diff
@@ -373,9 +380,12 @@ def parse_definitions(self, class_, all=False):
         while self.current is not None:
             self.log.debug("parsing definition list, current token is %r (%s)",
                            self.current.kind, self.current.value)
+            self.log.debug('got_newline: %s', self.stream.got_newline)
             if all and self.current.value == '__all__':
                 self.parse_all()
-            elif self.current.kind == tk.OP and self.current.value == '@':
+            elif (self.current.kind == tk.OP and
+                  self.current.value == '@' and
+                  self.stream.got_newline):
```
farmersez: I'm having trouble understanding why this works. As far as I understand, `stream.got_newline` only specifies if the previous token was a newline, not if the whole line was just an empty new line. Let's take your test case:
```python
(a
@b)
@foo
def bar():
    ...
```

If what I said is correct, how does this condition fails to detect `@b)` as another decorator? I'm guessing that it's something I didn't get about the way the parser works :^)

src/tests/parser_test.py
```diff
@@ -416,6 +417,46 @@ def test_raise_from():
...
+def test_simple_matrix_multiplication():
+    """Make sure 'a @ b' doesn't trip the parser."""
+    if sys.version_info.minor < 5:
+        return
+    parser = Parser()
+    code = CodeSnippet("""
+        def foo():
+            a @ b
+    """)
+    parser.parse(code, 'file_path')
```
Nurdok: I don't think this merits refactoring. It's the central action in the test.

src/pydocstyle/parser.py
```diff
@@ -271,7 +276,7 @@ class Parser(object):
     def parse(self, filelike, filename):
         """Parse the given file-like object and return its Module object."""
         # TODO: fix log
```
Nurdok: Done.

src/pydocstyle/parser.py
Nurdok: I changed the code to make this clearer, see if you get it now.

src/pydocstyle/parser.py
```diff
@@ -373,9 +380,12 @@ def parse_definitions(self, class_, all=False):
         while self.current is not None:
             self.log.debug("parsing definition list, current token is %r (%s)",
                            self.current.kind, self.current.value)
+            self.log.debug('got_newline: %s', self.stream.got_newline)
             if all and self.current.value == '__all__':
```
Nurdok: I agree, but not in this issue.

src/tests/parser_test.py
farmersez: The fact that `file_path` appears in every test is what bothers me, not the repetition of this call.

Nurdok: Allow me [to quote myself](http://amir.rachum.com/blog/2017/01/14/explicit-tests/):

> The thing is, while this string appears in many tests, it is an arbitrary string, which is coincidentally the same. We gain nothing from refactoring it out into a constant, while losing readability.

PR #252, opened 2017-04-26
Include documentation and tests in source distribution

tjanez:
This will make the source distribution on PyPI (e.g.: [pydocstyle-2.0.0.zip](https://pypi.python.org/packages/b0/33/52e2fa5662fbb2a8bc9bdfe31327cd58c2f856b31851b73fc5baf86a2a14/pydocstyle-2.0.0.zip#md5=2be57712e2743d23949bafed957d8c86)) complete and stand-alone.

Also, this would be very useful for distribution packages of pydocstyle since one could run the tests against the currently installed version of pydocstyle while building the package.

In addition, one could also build and include the documentation if desired.

For a wider perspective, see [my Fedora review request for pydocstyle](https://bugzilla.redhat.com/show_bug.cgi?id=1409654).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,9 @@
+# include meta data
 include README.rst LICENSE-MIT
+# include documentation and files needed by documentation
+recursive-include docs *.py *.rst
+include docs/Makefile docs/make.bat
+# include tests and files needed by tests
+include tox.ini
+recursive-include requirements *.txt
+recursive-include src/tests *.py
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -4,6 +4,15 @@ Release Notes
 **pydocstyle** version numbers follow the
 `Semantic Versioning <http://semver.org/>`_ specification.
 
+Current Development Version
+----------------------------
+
+Bug fixes
+
+* Include documentation and all files needed for building it in source
+  distribution.
+* Include tests and all files needed for running them in source distribution.
+
 2.0.0 - April 18th, 2017
 ------------------------
 


tjanez:
@sigmavirus24, thanks for looking into this pull request.

> This is simple to do, but it implies that the "source distribution" is more than just an installable package but instead a complete repository archive. Fedora, Debian, Gentoo, and other downstreams continuously try to make this argument and it falls down each time. What you all want is a source archive for a package, not a source distribution. Distributions should really just be the package and what's necessary to install it.

I know this is a controversial topic in the Python community as I've researched this myself for the projects I co-maintain ([Resolwe](https://github.com/genialis/resolwe), [Resolwe Bioinformatics](https://github.com/genialis/resolwe-bio), [Resolwe Python SDK](https://github.com/genialis/resolwe-bio-py), [Resolwe Runtime Utilities](https://github.com/genialis/resolwe-runtime-utils)).

Here is the summary of my research:

### What do Python packaging authorities recommend?

If you recall the [_Where should I put tests when packaging python modules?_ discussion on distutils mailing list in October 2015](https://mail.python.org/pipermail/distutils-sig/2015-October/thread.html#27003), there were quite different opinions from prominent Python developers:

- for example, **Donald Stufft** dislikes putting tests inside a package
  - "The supposed benefit is that anyone can run the tests at anytime, but I don't find that actually true because it means (as someone else pointed out) that you either have to depend on all your test dependencies or that there is already an additional step to install them. If you're going to have to locate and install the test dependencies, then you might as well fetch the tarball with tests as well."
  - "Some projects have test suites which are significantly large too. The PyCA cryptography project for instance, has to ship it's vectors as an additional package to reduce the size of the final build product."
  - https://mail.python.org/pipermail/distutils-sig/2015-October/027015.html

- on the other hand, **Barry Warsaw** is a big fan of putting the tests inside the package
  - "I've often looked at a package's tests to get a better understanding of something that was unclear from the documentation, or didn't work the way I expected.  Having the tests there in the installed package makes it easier to refer to."
  - "For complex packages with lots of subpackages, I have lots of tests/ directories, so that the unit tests are near to the code they test.  This way the source tree gets organized for free without additional complexity in an outside-the-package tests tree."
  - https://mail.python.org/pipermail/distutils-sig/2015-October/027044.html

- as a middle ground, **Ben Finney** prefers to have the tests under top-level tests/ directory and set up setup.py to ensure the tests are distributed with the source package, but not installed.
  - " I think this is a sensible approach:
    - The source package contains all the source files a developer would use to make further changes and test them.
    - The package for installation contains only those files useful for run-time users, plus metadata (e.g. copyright information)."
  - https://mail.python.org/pipermail/distutils-sig/2015-October/027005.html
  - https://mail.python.org/pipermail/distutils-sig/2015-October/027012.html
- **David Cournapeau** did a simple investigation on how many installed packages have tests installed:
  - "At Enthought, we support around 400 packages. If I install all our supported packages on Linux: I count ~ 308 packages by making the assumption that one directory in site-packages is a package (wrong but decent approximation)"
  - "I count as 'with tests' a package with at least one directory called test in say package"
  - "With those assumptions, I count **46 % packages with tests installed**. So it is not 'not very common'. Granted, we are biased toward scientific packages which to include tests in the package, but we also package popular web/etc... packages."
  - https://mail.python.org/pipermail/distutils-sig/2015-October/027032.html

### What do popular Python packages do?

- These Python packages **include tests in** the **source distribution**, **but exclude them from** the **binary package**:
    - virtualenv [[1]](https://github.com/pypa/virtualenv/blob/2db288fe1f8b6755f8df53b8a8a0667a0f71b1cf/setup.py#L120-L121) [[2]](https://github.com/pypa/virtualenv/blob/2db288fe1f8b6755f8df53b8a8a0667a0f71b1cf/MANIFEST.in#L2),
    - setuptools [[1]](https://github.com/pypa/setuptools/blob/44d2c01914baff076853f6402655589dde070703/setup.py#L101) [[2]](https://github.com/pypa/setuptools/blob/44d2c01914baff076853f6402655589dde070703/MANIFEST.in#L2-L3),
    - six [[1]](https://github.com/benjaminp/six/blob/3280aa8617b649a701705e4119a42ee18257eace/setup.py#L51)  [[2]](https://github.com/benjaminp/six/blob/3280aa8617b649a701705e4119a42ee18257eace/MANIFEST.in#L3),
    - boto [[1]](https://github.com/boto/boto/blob/de61c302917cfa4f30baf22699fcf6eacf65faf3/setup.py#L62-L85)  [[2]](https://github.com/boto/boto/blob/de61c302917cfa4f30baf22699fcf6eacf65faf3/MANIFEST.in#L11),
    - Flask [[1]](https://github.com/pallets/flask/blob/a34d0e6878c8c8a5fab05a69785c443f3c17075d/setup.py#L69)  [[2]](https://github.com/pallets/flask/blob/a34d0e6878c8c8a5fab05a69785c443f3c17075d/MANIFEST.in#L4),
    - Django [[1]](https://github.com/django/django/blob/58f02c498b659a906e9c30d946bd89bedc4717e5/setup.py#L44)  [[2]](https://github.com/django/django/blob/58f02c498b659a906e9c30d946bd89bedc4717e5/MANIFEST.in#L15),
    - Ansible [[1]](https://github.com/ansible/ansible/blob/d56ba0945dbf6c787a3e7f3e17b3db657b9dcee4/setup.py#L33) [[2]](https://github.com/ansible/ansible/blob/d56ba0945dbf6c787a3e7f3e17b3db657b9dcee4/MANIFEST.in#L15),
    - pycodestyle [[1]](https://github.com/PyCQA/pycodestyle/blob/ca5350195031c8e29c50292404eaec848bbb073b/setup.py#L33) [[2]](https://github.com/PyCQA/pycodestyle/blob/ca5350195031c8e29c50292404eaec848bbb073b/MANIFEST.in#L4).

- These packages **include tests in both source distrubtion and binary package**:
    - scikit-learn [[1]](https://github.com/scikit-learn/scikit-learn/blob/0.18.1/setup.py#L136),
    - simplejson [[1]](https://github.com/simplejson/simplejson/blob/c52efea57d0f733d28eb991eeebb121c53b3a48a/setup.py#L106),
    - pbr [[1]](http://git.openstack.org/cgit/openstack-dev/pbr/tree/setup.cfg#n28-n29).
  
- These packages **don't include tests in source distrubtion or binary package**:
    - pip [[1]](https://github.com/pypa/pip/blob/022248f6484fe87dc0ef5aec3437f4c7971fd14b/setup.py#L72) [[2]](https://github.com/pypa/pip/blob/022248f6484fe87dc0ef5aec3437f4c7971fd14b/MANIFEST.in),
    - requests [[1]](https://github.com/kennethreitz/requests/blob/b2289cd2d5d21bd31cf4a818a4e0ff6951b2317a/setup.py#L36-L47) [[2]](https://github.com/kennethreitz/requests/blob/b2289cd2d5d21bd31cf4a818a4e0ff6951b2317a/MANIFEST.in).

### Conclusion

Taking the approach of **including tests in** the **source distribution**, **but excluding them from** the **binary package** is reasonable, not against recommendations of Python packaging authorities and what a lot of popular Python packages do.

--------------------------------------------

@Nurdok, what's your take on this?

Nurdok:
@tjanez thanks for opening this. I read your research and various links and I'm not convinced either way yet. I think I'll think it over some more before I make up my mind.

sigmavirus24:
@tjanez here's the thing though, you're ignoring the difference between source _archives_ and source _distributions_. One (the archive) should include tests if it's provided, the other shouldn't. It's rather clearly defined in the packaging guide. If this project produced source archives without tests, I would absolutely agree. But it doesn't.

It's also worth mentioning that Barry and Ben are Debian and Ubuntu package maintainers. Downstream Linux distribution maintainers (not necessarily Barry and Ben) tend to go around complaining that tests aren't included in a source _distribution_ with little care for the distinction. Further, that discussion was a bit more generic. And I'll give you my *more* nuanced view of the matter.

People have different opinions of where to put tests. The two most common in the Python community are at the top level and inside the module next to the submodules they're testing. In the latter, tests are going to naturally be shipped with the package because they're inside the module directory. In that case, there's no argument as to whether or not to include them because every packaging library will include them automatically. In the former case, what's the actual value? The *vast majority* of source _distribution_ users are not packagers, they are people using pip without a binary package available for them. In that case, they're downloading a lot more data for no benefit. The minority of consumers of source _distributions_ are downstream redistributors who A) don't know about source _archives_, or B) don't have source _archives_ available to them. Those users are maybe 2% (probably less) of a maintainers source _distribution_ consumers. They're also the noisiest. No one installing from pip, conda, or any other tool benefit from the tests being in the package in either case. If there were user benefit for the additional bloat of test suites, I'd be absolutely very emphatic about every package including them by default.

Nurdok:
@sigmavirus24 @tjanez I am not familiar with the concept of source _archives_ in the Python packaging world. Is there a convention or platform to release those? Or does it just come down to downloading the repo's snapshot from GitHub?

sigmavirus24:
> Is there a convention or platform to release those?

They can be released to PyPI.

> Or does it just come down to downloading the repo's snapshot from GitHub?

This is frankly the same thing as a source archive with the exception that it includes CI service configuration and a probably other things that downstream redistributors don't care about.

Nurdok:
@sigmavirus24 Doesn't PyPI only accept source _distributions_, not _archives_?

sigmavirus24:
It will accept one or hte other at the moment, but not both. Archives can generate distributions for what it's worth.

jayvdb:
fwiw, I absolutely agree with this PR, as PyPI is the central auditable source of truth for any Python package, irrespective of where the source development & management is happening.  The tests are an intrinsic part of the software.  `setup.py test` is a default command for a reason, and running it without any tests in the source distribution will result in an erroneous zero exit code even though 0 tests were found thus none ran.  Many packages even include the tests under the package namespace.  CPython includes it as a separate module along side the others.

Most end users are fetching wheels when they are available, not source distributions/archives, so the size of the source tarball in pypi is hardly relevant.  And the size of the tests and docs makes it hardly worth the time to even discuss.  If someone does want the source, they are very likely expecting to get more than just the stuff needed to run it.  If they want to see the source, they should also be able to modify it, and refer to the docs, and to check their modifications they should be able to run the tests.

sigmavirus24:
I'd go so far as to say that any recent or modern Python project doesn't actually support `python setup.py test` unless they've been bullied into doing so and most times, that support bit rots to the point of the project removing it out of sheer frustration.

Saying "There's a standard" is one thing. But is it actually a standard if projects never actually use it or is it a defunct standard that people don't care about anymore. I'm genuinely convinced that it's the latter.

Sure some projects support it, but they're operating in different spaces than most. Finally, a source _archive_ which would include everything would satisfy what you all want. I still disagree with shoving unnecessary data into a source _distribution_, which is what's being requested here.

Nurdok:

Issue #245, opened 2017-04-06
Perform markup validation of docstrings, e.g. reStructuredText


peterjc:
This is an enhancement suggestion / wish list item.

I would like to be able to validate the markup used in Python docstrings, specifically as reStructuredText (aka RST, which as per PEP287 https://www.python.org/dev/peps/pep-0287/ is recommended), but others might want to use e.g. markdown.

Cross-reference ``restructuredtext-lint`` question https://github.com/twolfson/restructuredtext-lint/issues/34

peterjc:
As per the discussion on #254, the ``pydocstyle`` team currently would prefer this be a new plugin for ``flake8`` - see https://gitlab.com/pycqa/flake8/issues/337

(I propose to leave this issue open until such a plugin exists)

peterjc:
I've started working on this as a flake8 plugin under the name flake8-rst-docstrings,

https://github.com/peterjc/flake8-rst-docstrings

peterjc:
I've made the first release of the new plugin ``flake8-rst-docstrings`` on PyPI:

https://pypi.python.org/pypi/flake8-rst-docstrings

Unless there is a good place to mention this in the ``pydocstyle`` documentation, this issue can be closed.

PR #254, opened 2017-04-26
reStructuredText (RST) validation of docstrings

peterjc:
See #245. This is a proof of principle implementation to perform markup validation of docstrings as reStructuredText using docutils, based on public domain code by Todd Wolfson in his tool restructuredtext-lint ,
https://github.com/twolfson/restructuredtext-lint/blob/master/restructuredtext_lint/lint.py

Is something like this considered in scope for ``pydocstyle``?

If so, I'd personally prefer adding unique codes for each RST issue (perhaps numbered R### or D5## as a new class), rather than all under D415 as done here.

### DIFF
--- a/requirements/runtime.txt
+++ b/requirements/runtime.txt
@@ -1,1 +1,2 @@
 snowballstemmer==1.2.1
+docutils>=0.11,<1.0
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -13,7 +13,7 @@ from .config import IllegalConfiguration
 from .parser import (Package, Module, Class, NestedClass, Definition, AllError,
                      Method, Function, NestedFunction, Parser, StringIO,
                      ParseError)
-from .utils import log, is_blank, pairwise
+from .utils import log, is_blank, pairwise, rst_lint
 from .wordlists import IMPERATIVE_VERBS, IMPERATIVE_BLACKLIST, stem
 
 
@@ -641,6 +641,19 @@ class ConventionChecker(object):
                                      b is None)
             for err in self._check_section(docstring, definition, new_ctx):
                 yield err
+
+    @check_for(Definition)
+    def check_valid_rst(self, definition, docstring):
+        """D415: Fails reStructuredText validation.
+
+        PEP-287 recommends reStructuredText (RST) as the docstring markup
+        format, and this checks the docstring passes validation as RST.
+
+        Extract the docstring, remove the indentation as per the trim
+        function define in PEP-257, then validate it as RST.
+        """
+        if docstring:
+            return [violations.D415(_) for _ in rst_lint(docstring)]
 
 
 parse = Parser()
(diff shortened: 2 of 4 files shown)


sigmavirus24:
I'm not convinced this needs to be part of pydocstyle. I think this could handily live as a Flake8 extension which gives the users the choice to pick one that validates Markdown. As usual, though, I defer the actual decision to @Nurdok

peterjc:
If flake8 is deemed a more suitable home, that works for me. But as this is docstring-centric I'm asking here first.

sigmavirus24:
> If flake8 is deemed a more suitable home, that works for me.

I seem to have communicated poorly. I think this should be an extension that can be run independent of pydocstyle by making it a plugin for Flake8, not core to Flake8.

peterjc:
@Nurdok where would you prefer to see reStructuredText validation of Python docstrings?

- as a separate self-contained plugin for ``flake8`` as suggested by @sigmavirus24,
- as part of ``pydocstyle`` (as in this proof of principle, which I assume would automatically become available in ``flake8`` via the ``flake8-docstrings`` plugin)

Speaking as an end-user, I would be happy using the functionality either way.

Nurdok:
@peterjc thanks for the PR! I admit I'm not sure where I stand on this. I am a little bit hesitant to add an error code that essentially can result from an unknown number of errors and reasons. To avoid feature creep, I prefer to not merge this for now, but I am very open to discussion about it.

peterjc:
Assigning everything to a single code (initially D415) was for simplicity as a proof of principle.

I suspect that in a similar way to how flake8 assigned its own numbers to the pyflakes checks, we might want to assign numbers to the docutils RST checks. These could be D501, D502, etc within pydocstyle or a whole new class like R101 (R for RST or reStructuredText) if that is available in the flake8 ecosystem.

sigmavirus24:
@peterjc have you heard of doc8? I think this might be a good addition to that tool. It specifically checks for rST errors

peterjc:
I'd not heard of doc8, although like ``restructuredtext-lint`` it seems to be focused on linting ``*.rst`` files?

https://pypi.python.org/pypi/doc8
https://launchpad.net/doc8
https://github.com/openstack/doc8

Nurdok:
Right now, it looks like creating a separate plugin for `flake8` is my preferred solution.

peterjc:
OK, I've logged https://gitlab.com/pycqa/flake8/issues/337 to try to get a little guidance from the ``flake8`` team before doing any more work on this (e.g. naming recommendations).

peterjc:
I've started working on this as a flake8 plugin under the name flake8-rst-docstrings,

https://github.com/peterjc/flake8-rst-docstrings

peterjc:
I've made the first release of the new plugin ``flake8-rst-docstrings`` on PyPI:

https://pypi.python.org/pypi/flake8-rst-docstrings

(And I will stop commenting here now)

Issue #256, opened 2017-05-08
--source is broken


brbsix:
The `--source` option is broken. Output looks like:

	src/pydocstyle/violations.py:87 in public method `__lt__`:
		    D105: Missing docstring in magic method

	876:     def __lt__(self, other):
		87:     def __lt__(self, other):
	886:         return (self.filename, self.line) < (other.filename, other.line)
		88:         return (self.filename, self.line) < (other.filename, other.line)

When it (presumably) should look like:

	src/pydocstyle/violations.py:87 in public method `__lt__`:
		    D105: Missing docstring in magic method

	87:     def __lt__(self, other):
	88:         return (self.filename, self.line) < (other.filename, other.line)

The issue originates in `Errors.lines`, found here: https://github.com/PyCQA/pydocstyle/blob/master/src/pydocstyle/violations.py#L49

It looks like some code may have been committed accidentally before it was complete.

Nurdok:
Fixed by #260 .

PR #257, opened 2017-05-08
Fix --source option

brbsix:
This fixes the erroneous output described in https://github.com/PyCQA/pydocstyle/issues/256.

Perhaps someone else can decide if the line numbers displayed by `--source` should be of fixed width (in which case lines https://github.com/PyCQA/pydocstyle/blob/master/src/pydocstyle/violations.py#L56-L59 can be removed) or not (in which case line https://github.com/PyCQA/pydocstyle/blob/master/src/pydocstyle/violations.py#L60 can be removed).


Nurdok:
Can you please add some tests to make sure this doesn't break again?

Nurdok:
Closing - continued in #260.

Issue #198, opened 2016-07-14
Separate error for missing docstring on nested (Meta) classes


hvdklauw:
In django the usage of a nested Meta class is pretty common, however adding docstrings to those classes is just silly.

Would be nice to have it separated from normal classes so we could turn off the need for a docstring there.

jayvdb:
This issue is about a nested class named `Meta`.
https://docs.djangoproject.com/en/dev/topics/db/models/#meta-options

I doubt special-casing that name is appropiate, although it could be done in conjunction with special casing of the outer-class subclassing `models.Model`, but then we have the problem that the ast cant detect indirect subclassing.

The other approach is a separate error for any nested classes, and probably also nested functions if they are not already a different error or no error.

Any other approaches?

hvdklauw:
I dislike the first option, then soon there will be some other special case somewhere in the python eco system and we'll end up with many lines of code to handle all those special cases.

I think the separate errors for nested classes and for nested functions (2 extra errors) would be much nicer.

thejcannon:
@Nurdok this can be closed, right? As #261 has added D106 for this case?

PR #261, opened 2017-05-30
Mark missing docstrings for nested classes as violation D106

rhoog:
Currently D101 covers both classes and nested classes. This is a problem if you want to exclude e.g. a Meta nested class from needing a docstring.
In this pull request I put Nested Classes that miss a docstring under D106. This allows me to configure pydocstyle to check for D101, but exclude the D106 situation.

This problem is also discussed at: https://github.com/PyCQA/pydocstyle/issues/198

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -4,6 +4,14 @@ Release Notes
 **pydocstyle** version numbers follow the
 `Semantic Versioning <http://semver.org/>`_ specification.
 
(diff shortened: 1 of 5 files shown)


rhoog:
@Nurdok Sure. I've assumed v2.0.1 is up next and put it under this version in the release docs.

Nurdok:
Merged via #263. Thanks, @rhoog!

Issue #195, opened 2016-06-29
support stdin


dirk-thomas:
If `pydocstyle` could be invoked with `-` to indicate to read the content which should be linted from `stdin` that would enable better integration of the tool in various editors. E.g. on-the-fly linting (with now need to save the file explicitly) in Atom: AtomLinter/linter-pydocstyle#34

sigmavirus24:
@dirk-thomas flake8-docstrings + flake8 make this possible and easy (for a short-ish term solution).

milkey-mouse:
Where would `pydocstyle` look for config files? Would it just use the defaults?

Nurdok:
It should not load configuration files unless explicitly passed with #117.

PR #265, opened 2017-06-17
Support stdin via hyphen as filename.

squdle:
Originally, I was looking to use pydocstyle in vim without using other plugins, and couldn't find a good way to pipe stdin. Eventually noticed that this was raised already in [issues/195](https://github.com/PyCQA/pydocstyle/issues/195). Proposed behavior is to use `-` as a 'filename' argument to notify that stdin is being piped. This might be too hacky, feel free to decline if this is the case!


Nurdok:
Hey @squdle, just a friendly ping.

Nurdok:

Issue #273, opened 2017-07-30
Allow __init__ to not have a docstring


mitar:
I prefer to have docstring for __init__ as part of the class docstring. Currently it seems there is no way to disable just the requirement for the __init__ docstring.

sigmavirus24:
`__init__` should be considered a "magic" method and thus should be reported as D105. You can ignore D105 altogether to get rid of that.

mitar:
Currently it is ignored [as a magic method](https://github.com/PyCQA/pydocstyle/blob/master/src/pydocstyle/parser.py#L208). But yes, as a workaround I patched `VARIADIC_MAGIC_METHODS` to not contain `__init__`.

But because it seems there are different ways people would like this, maybe for `__init__` there could be a configuration.

Nurdok:
I think this is common enough to merit a dedicated `D1xx` error code.

AndrewPashkin:
Numpy docstring guide specifically [says][1]:
>  The constructor (__init__) should also be documented here [in a class docstring]... 

[1]: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#id7

Nurdok:
@AndrewPashkin This is already merged :)

AndrewPashkin:
@Nurdok, the merge request just created a dedicated error code for that. But Numpy rules state that `__init__` SHOULD be documented in the class docstring and so it's not an error for `__init__` to not have a docstring.

mitar:
Yes, so people who are using Numpy rules can configure pydocstyle to ignore that dedicated error. I think requiring D101 also requires one to document constructor in class docstring, no?

AndrewPashkin:
But in Numpy rules it's not an error at all.

Nurdok:
Seems to me that if this is the case, it's a simple matter of removing D107 from being on by default in the `numpy` convention.

PR #277, opened 2017-08-28
Introduce D107 code for missing __init__ docstring

tjanson:
As described in #273, it would be useful to have a separate error code for missing docstrings on `__init__` methods (which currently fall under `D102`).
This PR introduces `D107` for this purpose.

This enables the user to use only class docstrings (as described by [this Google style](https://github.com/sphinx-doc/sphinx/blob/1.6.3/doc/ext/example_google.py#L201-L205)) by ignoring `D107`.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -12,6 +12,8 @@ New Features
 
 * Public nested classes missing a docstring are now reported as D106 instead
   of D101 (#198, #261).
+* ``__init__`` methods missing a docstring are now reported as D107 instead of
+  D102 (#273).
 * Added support for Python 3.6 (#270).
 
 Bug Fixes
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -13,7 +13,7 @@ New Features
 * Public nested classes missing a docstring are now reported as D106 instead
   of D101 (#198, #261).
 * ``__init__`` methods missing a docstring are now reported as D107 instead of
-  D102 (#273).
+  D102 (#273, #277).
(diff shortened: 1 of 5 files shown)


Nurdok:
The code itself looks good to me and I am inclined to merge this, but I have a thought - isn't the same reasoning for having a separate error code for `__init__` also applies to the other variadic magic methods (i.e., `__call__` and `__new__`)? For example, I can also see how `__call__` would be documented in the class docstring instead of in the `__call__` method docstring.

So, I am considering applying this to all variadic magic methods and changing the error description accordingly. Any thoughts?

/cc @FarmerSez

FarmerSez:
I tend to disagree @Nurdok .
`__init__` always constructs an object, and usually they way it's done is pretty boring to document.
If someone had overridden `__call__` or `__new__`, something interesting is going on, and I'd like to see documentation for that wizardry within the method. That doesn't mean that there shouldn't be any general documentation to that behavior in the class's docstring - there should, but I expect more info within the method.

For example, you know how django models work. Wouldn't you like to see a docstring for [that](https://github.com/django/django/blob/master/django/db/models/base.py) `__new__` method? (line 73)

tjanson:
Yeah, I agree @FarmerSez. I definitely think `__init__` should be treated separately, by itself.

It certainly doesn't hurt to have an extra code for the remaining variadic magic methods, but I personally don't need that. If I were you I'd handle it just like this case: Accept a PR if someone comes along and wants that feature, but don't (pro-)actively implement it.

PR #280, opened 2017-09-03
Config file warning

FarmerSez:
Thanks for submitting a PR!


src/pydocstyle/config.py
```diff
@@ -256,10 +256,17 @@ def _get_config(self, node):
...
+
             if None in self._cache:
                 return self._cache[None]
             options, _ = self._read_configuration_file(self._run_conf.config)
-            config = self._create_check_config(options)
+
+            if options is None:
+                log.warning('Configuration file does not contain a '
+                            'pydocstyle section. Using default configuration.')
+                config = self._create_check_config(self._options)
```
Nurdok: This function is called for every directory in a project, so a lack of a `pydocstyle` section should generally be treated by just skipping it and checking the parent directory. If you use the default config, you might screw up the configuration (this might happen if, for example, there's a `setup.cfg` file for a different tool).

This logic should only be applied if `self._run_config.config` is not `None`, i.e., if a specific config file is specified.

Nurdok: Sorry, I thought this block was in the top level of the method, but in fact it's inside the `self._run_config.config` check. But add a test for this anyway (a config file in the hierarchy that doesn't have a correct section name).

FarmerSez: 😸

Issue #250, opened 2017-04-25
Multi-line entries in configuration file


peterjc:
Related tool ``flake8`` explicitly supports multi-line configuration entries (which can be intermixed with comments) as described in http://flake8.pycqa.org/en/latest/user/configuration.html

e.g.

```
[flake8]
# it's not a bug that we aren't using all of hacking
ignore =
    # F812: list comprehension redefines ...
    F812,
    # H101: Use TODO(NAME)
    H101,
    # H202: assertRaises Exception too broad
    H202,
    # H233: Python 3.x incompatible use of print operator
    H233,
    # H301: one import per line
    H301,
    # H306: imports not in alphabetical order (time, os)
    H306,
    # H401: docstring should not start with a space
    H401,
    # H403: multi line docstrings should end on a new line
    H403,
    # H404: multi line docstring should start without a leading new line
    H404,
    # H405: multi line docstring summary not separated with an empty line
    H405,
    # H501: Do not use self.__dict__ for string formatting
    H501
```

I would like to be able to do the same in ``.pydocstyle`` for our ``ignore =`` list in order to comment why we ignore some entries, and facilitate merging edits to the file.

In my testing with pydocstyle v2.0.0 an example like this fails (it seems to use an empty ignore list).

peterjc:
For anyone else with a similar use case, I decided to stop running ``pydocstyle`` directly, and instead use it via ``flake8`` with the ``flake8-docstring`` plugin. This means I can put the D### codes into a ``.flake8`` configuration file.

FarmerSez:
Done in #281 🖖

PR #281, opened 2017-09-08
#250 - Added support for multi-lined config entries

FarmerSez:
Thanks for submitting a PR!

Solves #250 .

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -17,6 +17,7 @@ New Features
 * Added support for Python 3.6 (#270).
 * Specifying an invalid error code prefix (e.g., ``--select=D9``) will print
   a warning message to ``stderr`` (#253, #279).
+* Configuration files now support multiple-lined entries (#250, #281).
 
 Bug Fixes
 
--- a/requirements/runtime.txt
+++ b/requirements/runtime.txt
@@ -1,1 +1,2 @@
 snowballstemmer==1.2.1
+configparser==3.5.0
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
-import sys
 from __future__ import with_statement
 from setuptools import setup
+import sys
 
 # Do not update the version manually - it is managed by `bumpversion`.
 version = '2.0.1rc'
--- a/setup.py
+++ b/setup.py
@@ -6,15 +6,15 @@ import sys
 version = '2.0.1rc'
 
 
-REQUIRES = [
-        'snowballstemmer',
-        'six',
-    ]
+requirements = [
+    'snowballstemmer',
+    'six',
+]
 
 
 # Python3 to Python2 backport support.
 if sys.version_info[0] == 2:
-    REQUIRES.append('configparser == 3.5.0')
+    requirements.append('configparser')
 
 
 setup(
@@ -38,7 +38,7 @@ setup(
     packages=('pydocstyle',),
     package_dir={'': 'src'},
     package_data={'pydocstyle': ['data/*.txt']},
-    install_requires=REQUIRES,
+    install_requires=requirements,
     entry_points={
         'console_scripts': [
             'pydocstyle = pydocstyle.cli:main',
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,21 @@
+import sys
 from __future__ import with_statement
 from setuptools import setup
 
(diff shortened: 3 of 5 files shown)


src/pydocstyle/config.py
```diff
@@ -7,9 +7,9 @@
 from re import compile as re
 
 
-try:  # Python 3.x
+try:  # Python 2.x
     from ConfigParser import RawConfigParser
```
sigmavirus24: Honestly, I would just use the backport from Python 3. The Python 2 version fo ConfigParser is *awful*.

src/pydocstyle/config.py
```diff
@@ -453,6 +453,12 @@ def _expand_error_codes(code_parts):
 
         try:
             for part in code_parts:
+                # Dealing with split-lined configurations. The part might begin
+                # with a whitespace and/or a comment marker.
+                part = part.strip()
+                if not part or part.startswith(('#', ';')):
```
sigmavirus24: For example, if you use the backport, this goes away.

src/pydocstyle/config.py
FarmerSez: @Nurdok , what say you?

Nurdok: Since we crossed to chasm of package dependencies, I think you should use whichever package is best for the job, even if it adds a dependency.

requirements/runtime.txt
```diff
@@ -1 +1,2 @@
 snowballstemmer==1.2.1
+configparser==3.5.0
```
Nurdok: You also need to add this to `setup.py`, only for Python 2.x.

src/tests/test_integration.py
```diff
@@ -277,6 +277,29 @@ def foo():
...
+        example.write(textwrap.dedent("""\
+            class Foo(object):
+                "Doc string"
+                def foo():
+                    pass
+        """))
+
+    select_string = ('D100,\n'
+                     '  #D103,\n'
+                     ' D204, D300')
```
Nurdok: Also add a line that has in-line comment, such as:
```
D301,  # we want to select this
D302
```

setup.py
```diff
@@ -28,6 +28,7 @@
     install_requires=[
         'snowballstemmer',
         'six',
+        'configparser;python_version<"3.3"',
```
sigmavirus24: This will break on old enough versions of pip and setuptools which, if I remember correctly, is anyone using Ubuntu between 14.04 and 16.04

FarmerSez: @Nurdok and I were just discussing it. I saw [this](https://hynek.me/articles/conditional-python-dependencies/) and thought about doing it in the `setup.py`, but it's reeeeeeeeeeeeally gross.  Have you encountered that issue before and have an example / boilerplate? 😜

FarmerSez: Bump, @sigmavirus24 ?

setup.py
```diff
@@ -1,9 +1,22 @@
 from __future__ import with_statement
 from setuptools import setup
+import sys
 
 # Do not update the version manually - it is managed by `bumpversion`.
 version = '2.0.1rc'
 
+
+REQUIRES = [
+        'snowballstemmer',
```
Nurdok: Indentation seems weird here.
Should be:
```.py
REQUIRES = [
    'snowballstemmer',
    'six',
]

Comment (Nurdok) [reply setup.py]: Also, why is `REQUIRED` capital cased? It's not a constant.

File: src/tests/test_integration.py [hunk 7]
```diff
@@ -277,6 +277,29 @@ def foo():
     assert 'file does not contain a pydocstyle section' not in err
 
 
+def test_multiple_lined_config_file(env):
+    """Blah."""
```
Nurdok: Oops :)

docs/release_notes.rst
```diff
@@ -17,6 +17,7 @@ New Features
 * Added support for Python 3.6 (#270).
 * Specifying an invalid error code prefix (e.g., ``--select=D9``) will print
   a warning message to ``stderr`` (#253, #279).
+* Configuration files now support multiple-lined entries (#250, #281).
```
peterjc: I would suggest "multiple line entries" here (not line*d*).

Issue #231, opened 2017-01-22
Improve description of --add-ignore option


cdeil:
I think the current `--add-ignore` option is a copy of `--add-select`, and it should be changed from "errors to check" to "errors to ignore", no?
```
  --add-select=<codes>  amend the list of errors to check for by specifying
                        more error codes to check.
  --add-ignore=<codes>  amend the list of errors to check for by specifying
                        more error codes to ignore.
```

An unrelated suggestion concerning:
```
$ pydocstyle --convention=numpy --ignore=D410
Cannot pass both ignore and convention. They are mutually exclusive.
```
I would have found it useful if that message had this at the end:
```
You can use `--add-ignore` instead of `--ignore`.
```

cdeil:
Is it possible to select just one given code with the number convention?

This doesn't work:
```
$ pydocstyle --select=D412 --convention=numpy
Cannot pass both select and convention. They are mutually exclusive.
```

Nurdok:
You can do `pydocstyle --convention=numpy --add-select=D412`.

cdeil:
@Nurdok - What `pydocstyle --convention=numpy --add-select=D412` does is to add D412 to a list of default codes, i.e. in the report I get D412 but a lot of other things as well.

I want only `D412`, no other codes.

Nurdok:
Oh, so just do `pydocstyle --select=D412`.

Nurdok:
Looking back at this, it looks like the flag description is precise, but perhaps non-intuitive:

```
  --select=<codes>      choose the basic list of checked errors by specifying
                        which errors to check for (with a list of comma-
                        separated error codes or prefixes). for example:
                        --select=D101,D2
  --ignore=<codes>      choose the basic list of checked errors by specifying
                        which errors to ignore (with a list of comma-separated
                        error codes or prefixes). for example:
                        --ignore=D101,D2
  --convention=<name>   choose the basic list of checked errors by specifying
                        an existing convention. Possible conventions: pep257,
                        numpy
  --add-select=<codes>  amend the list of errors to check for by specifying
                        more error codes to check.
  --add-ignore=<codes>  amend the list of errors to check for by specifying
                        more error codes to ignore.
```
The phrasing was done to emphasize the point that the `--add-XXX` flags only _amend_ the list. I'm not sure what a better per-flag description can be, or how to make this more clear. Perhaps we just need to add a longer, general description of error code selection in the `--help` text?

sigmavirus24:
[...]

PR #283, opened 2017-09-22
Better help section

FarmerSez:
Thanks for submitting a PR!

Solves #231

### DIFF
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -542,11 +542,20 @@ class ConfigurationParser(object):
     @classmethod
     def _create_option_parser(cls):
         """Return an option parser to parse the command line arguments."""
-        from optparse import OptionParser
+        from optparse import OptionParser, OptionGroup
+
+        epilog = ('How Error Code Selection Works: '
+                  'There are 3 options that decide which error codes will be '
+                  'tested: --convention, --select and --ignore. Only one '
+                  '(or none) of these three options can be used '
+                  '(when none of them was selected, the default value is '
+                  '`--convention=pep257`).'
+                  'These options select the initial lise of errors to check.')
 
         parser = OptionParser(
             version=__version__,
-            usage='Usage: pydocstyle [options] [<file|dir>...]')
+            usage='Usage: pydocstyle [options] [<file|dir>...]',
+            epilog='')
 
         option = parser.add_option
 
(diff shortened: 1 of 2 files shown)


FarmerSez:
This is how it looks now:

```
(pydocstyle) C:\Users\shach\code\pydocstyle\src>python pydocstyle --help
Usage: pydocstyle [options] [<file|dir>...]

Options:
  --version             show program's version number and exit
  -h, --help            show this help message and exit
  -e, --explain         show explanation of each error
  -s, --source          show source for each error
  -d, --debug           print debug information
  -v, --verbose         print status information
  --count               print total number of errors to stdout
  --config=<path>       use given config file and disable config discovery
  --match=<pattern>     check only files that exactly match <pattern> regular
                        expression; default is --match='(?!test_).*\.py' which
                        matches files that don't start with 'test_' but end
                        with '.py'
  --match-dir=<pattern>
                        search only dirs that exactly match <pattern> regular
                        expression; default is --match-dir='[^\.].*', which
                        matches all dirs that don't start with a dot
  --ignore-decorators=<decorators>
                        ignore any functions or methods that are decorated by
                        a function with a name fitting the <decorators>
                        regular expression; default is --ignore-decorators=''
                        which does not ignore any decorated functions.

  Error Check Options:
    Only one of --select, --ignore or --convention can be specified. If
    none is specified, defaults to `--convention=pep257`. These three
    options select the "basic list" of error codes to check. If you wish
    to change that list (for example, if you selected a known convention
    but wish to ignore a specific error from it or add a new one) you can
    use `--add-[ignore/select]` in order to do so.

    --select=<codes>    choose the basic list of checked errors by specifying
                        which errors to check for (with a list of comma-
                        separated error codes or prefixes). for example:
                        --select=D101,D2
    --ignore=<codes>    choose the basic list of checked errors by specifying
                        which errors to ignore out of all of the available
                        error codes (with a list of comma-separated error
                        codes or prefixes). for example: --ignore=D101,D2
    --convention=<name>
                        choose the basic list of checked errors by specifying
                        an existing convention. Possible conventions: pep257,
                        numpy.
    --add-select=<codes>
                        add extra error codes to check to the basic list of
                        errors previously set by --select, --ignore or
                        --convention.
    --add-ignore=<codes>
                        ignore extra error codes by removing them from the
                        basic list previously set by --select, --ignore or
                        --convention.
```

src/pydocstyle/config.py
```diff
@@ -303,8 +303,12 @@ def _read_configuration_file(self, path):
         should_inherit = True
 
         if parser.read(path) and self._get_section_name(parser):
+            all_options = [o for o in self._parser.option_list]
```
Nurdok: Can't this be just `all_options = self_parser.option_list[:]`?

FarmerSez: But mom, all the cool kids are doing it!

PR #310, opened 2018-04-24
WIP: run checks in parallel

jobevers:
Using multiprocessing.Pool to run tests in parallel. For me, it cut the time to run pydocstyle from 30 seconds to 5.

WIP because I don't think this should be the default mode, but wasn't sure what would be good configuration/command line flags to use.


sigmavirus24:
I wonder if we shouldn't try tearing out this logic from Flake8 so it could be reused here

Nurdok:

Issue #311, opened 2018-04-30
Incorrect D405 on lowercase "attributes" used in docstring.


TimZaman:
```py
    def clip(self, attributes_per_polyline):  # noqa: D405 (pydocstring bug)
        """
        Some stuff here.

        Args:
            attributes_per_polyline (tensor): attributes for each polyline.

        Returns:
            stuff
        """
```

Throws: 
`D405`
```
stuff.py:1337 in public method `clip`:
        D405: Section name should be properly capitalized ('Attributes', not 'attributes')
```
I think 'attributes' is seen as some kind of special keyword here by pydocstyle, which it isn't/doesn't have to be.

shacharoo:
'Attributes' is a section name in the numpy docstring conventions. Although this is a bug (we recognize the prefix "attributes" of "attributes_per_polyline"), you should make sure you're using the right conventions for your project.

TimZaman:
Our convention is Args.

PR #317, opened 2018-06-06
Fix false-positive D405 recognition

shacharoo:
Solves #311.

The problem was that the regex split words when reached an underscore, so "arguments_bla" was caught as "arguments".

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -480,10 +480,7 @@ class ConventionChecker(object):
             context.section_name.strip()
         ).strip()
 
-        section_suffix_is_only_colon = (
-            section_name_suffix.startswith(':') and
-            is_blank(section_name_suffix[1:])
-        )
+        section_suffix_is_only_colon = section_name_suffix == ':'
 
         punctuation = [',', ';', '.', '-', '\\', '/', ']', '}', ')']
         prev_line_ends_with_punctuation = \
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -436,6 +436,7 @@ class ConventionChecker(object):
 
         For example, if `line` is "  Hello world!!!", returns "Hello world".
         """
+        result = re("[\w ]+").match(line.strip())
         if result is not None:
             return result.group()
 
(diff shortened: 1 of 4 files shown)


src/pydocstyle/checker.py
```diff
@@ -454,30 +454,42 @@ def _is_a_docstring_section(context):
...
+            * There's no suffix to the section name or it's just a colon AND
+            * The previous line is empty OR it ends with punctuation.
 
         If one of the conditions is true, we will consider the line as
         a section name.
         """
-        section_name_suffix = context.line.lstrip(context.section_name).strip()
+        section_name_suffix = context.line.strip().lstrip(
+            context.section_name.strip()
+        ).strip()
```
Nurdok: Weird formatting, can you make it look better?

src/pydocstyle/checker.py
```diff
@@ -454,30 +454,42 @@ def _is_a_docstring_section(context):
...
+        section_suffix_is_only_colon = section_name_suffix == ':'
 
         punctuation = [',', ';', '.', '-', '\\', '/', ']', '}', ')']
         prev_line_ends_with_punctuation = \
             any(context.previous_line.strip().endswith(x) for x in punctuation)
 
-        return (is_blank(section_name_suffix) or
-                prev_line_ends_with_punctuation or
-                is_blank(context.previous_line))
+        return ((is_blank(section_name_suffix) or
```
Nurdok: If we're going verbose 😄:
```.py
this_line_looks_like_a_section_name = is_blank(section_name_suffix) or prev_line_ends_with_punctuation
prev_line_looks_like_end_of_paragraph = prev_line_ends_with_punctuation or is_blank(context.previous_line)
return this_line_looks_like_a_section_name and prev_line_looks_like_end_of_paragraph
```

src/tests/test_cases/sections.py
```diff
@@ -215,3 +215,28 @@ def multiple_sections():
...
     """
+
+
+@expect(_D213)
+def false_positive_section_prefix():
+    """Toggle the gizmo.
+
+    Parameters
+    ----------
+    attributes_are_fun: attributes for the function.
```
Nurdok: Perhaps also add a case where the parameter name is _exactly_ a section header name (e.g., `attributes`)?

shacharoo: Look at the test below it - it's exactly that.

src/pydocstyle/checker.py
shacharoo: I don't know if you're sarcastic, but I actually think it's better!

PR #331, opened 2018-09-27
Added PEP526-Styled ```__all__``` Support

DevynCJohnson:
PyDocStyle fails to run if the file it scans uses an ```__all__``` that is structured like ```__all__: list = [...]```. Since variable annotations ([PEP526](https://www.python.org/dev/peps/pep-0526/)) are permitted in Python, PyDocStyle should not fail to run due to an annotated ```__all__```.

### DIFF
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -446,9 +446,7 @@ class Parser(object):
             if self.current.value != r'list':
                 raise AllError(r'Could not evaluate contents of __all__. ')
             self.consume(tk.NAME)
-            if self.current.value != r'=':
-                raise AllError(r'Could not evaluate contents of __all__. ')
-        elif self.current.value != r'=':
+        if self.current.value != r'=':
             raise AllError(r'Could not evaluate contents of __all__. ')
         self.consume(tk.OP)
         if self.current.value not in r'([':
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -439,13 +439,20 @@ class Parser(object):
 
     def parse_all(self):
         """Parse the __all__ definition in a module."""
-        assert self.current.value == '__all__'
+        assert self.current.value == r'__all__'
         self.consume(tk.NAME)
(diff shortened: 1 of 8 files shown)


src/pydocstyle/parser.py
```diff
@@ -439,13 +439,20 @@ def parse_definitions(self, class_, all=False):
 
     def parse_all(self):
         """Parse the __all__ definition in a module."""
-        assert self.current.value == '__all__'
+        assert self.current.value == r'__all__'
```
Nurdok: Why use raw strings everywhere?

src/pydocstyle/parser.py
```diff
@@ -439,13 +439,20 @@ def parse_definitions(self, class_, all=False):
...
-            raise AllError('Could not evaluate contents of __all__. ')
+        if self.current.value == r':':  # PEP526-Styled __all__
+            self.consume(tk.OP)
+            if self.current.value != r'list':
+                raise AllError(r'Could not evaluate contents of __all__. ')
+            self.consume(tk.NAME)
+            if self.current.value != r'=':
+                raise AllError(r'Could not evaluate contents of __all__. ')
+        elif self.current.value != r'=':
+            raise AllError(r'Could not evaluate contents of __all__. ')
```
Nurdok: These can extracted outside the `if`, something like this:
```.py
if self.current.value == r':':  # PEP526-Styled __all__
    ...
if self.current.value != r'=':
    raise AllError(r'Could not evaluate contents of __all__. ')

```

src/pydocstyle/parser.py
DevynCJohnson: Raw strings are processed slightly faster than plain-strings. There are no escapes in my strings, so I can make it a raw string so that the parser does not scan the strings for character escapes.

samj1912:
Sorry about the late response. The PR looks good but the tests seem to be failing, possibly because the branch is not updated. Can you please rebase the changes on top of master?

samj1912:

PR #342, opened 2018-10-14
Introducing type hints

shacharoo:
Added  the minimal code needed for typehint validation.
I also removed some whitespaces in tox.ini - it seems that tox doesn't remove them (for instance, py34 install env was called `py34- install`).

### DIFF
--- a/src/pydocstyle/cli.py
+++ b/src/pydocstyle/cli.py
@@ -49,7 +49,7 @@ def run_pydocstyle():
         return ReturnCode.invalid_options
 
     count = 0
-    for error in errors:
+    for error in errors:  # type: ignore
         if hasattr(error, 'code'):
             sys.stdout.write('%s\n' % error)
         count += 1
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -121,7 +121,7 @@ class Definition(Value):
 class Module(Definition):
     """A Python source code module."""
 
-    _fields = ('name', '_source', 'start', 'end', 'decorators', 'docstring',
+    _fields = ('name', '_source', 'start', 'end', 'decorators', 'docstring',  # type: ignore
                'children', 'parent', '_dunder_all', 'dunder_all_error',
                'future_imports', 'skipped_error_codes')
     _nest = staticmethod(lambda s: {'def': Function, 'class': Class}[s])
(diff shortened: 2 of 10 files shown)


src/pydocstyle/cli.py
```diff
@@ -49,7 +49,7 @@ def run_pydocstyle():
         return ReturnCode.invalid_options
 
     count = 0
-    for error in errors:
+    for error in errors:  # type: ignore
```
Nurdok: Why?

src/pydocstyle/parser.py
```diff
@@ -102,7 +102,7 @@ def __str__(self):
 class Module(Definition):
     """A Python source code module."""
 
-    _fields = ('name', '_source', 'start', 'end', 'decorators', 'docstring',
+    _fields = ('name', '_source', 'start', 'end', 'decorators', 'docstring',  # type: ignore
```
Nurdok: Why?

src/tests/test_cases/sections.py
```diff
@@ -48,7 +48,7 @@ def no_underline():
 
 @expect(_D213)
 @expect("D407: Missing dashed underline after section ('Returns')")
-def no_underline():
+def no_underline_and_no_description():
```
Nurdok: Why?

src/pydocstyle/parser.py
shacharoo: Overriding the `_fields` member of the base class:
```
py37-tests runtests: commands[1] | mypy --config-file=tox.ini src/
src\pydocstyle\parser.py:105: error: Incompatible types in assignment (expression has type "Tuple[str, str, str, str, str, str, str, str, str, str, str, str]", base class "Definition" defined the type as "Tuple[str, str, str, str, str, str, str, str, str]")
```

src/pydocstyle/cli.py
shacharoo: Hiding the `error` variable defined above in `except IllegalConfiguration as error:`. I could rename it if you want.

src/tests/test_cases/sections.py
shacharoo: A function with the same name was defined above, my (historical) bad 😋

src/pydocstyle/cli.py
shacharoo:
```
py37-tests runtests: commands[1] | mypy --config-file=tox.ini src/
src\pydocstyle\cli.py:52: error: Assignment to variable 'error' outside except: block
```

PR #344, opened 2018-10-20
Type Hints!

Nurdok:
Adding type hints to the codebase.

### DIFF
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -4,3 +4,4 @@ mock==2.0.0
 pathlib
 mypy==0.630
 six
+typing
--- a/src/pydocstyle/violations.py
+++ b/src/pydocstyle/violations.py
@@ -1,9 +1,12 @@
 """Docstring violation definition."""
+
 from itertools import dropwhile
 from functools import partial
 from collections import namedtuple
+from typing import Iterable, Optional, List, Callable, Any
 
 from .utils import is_blank
+from .parser import Definition
 
 
 __all__ = ('Error', 'ErrorRegistry', 'conventions')
@@ -19,7 +22,13 @@ class Error:
     explain = False
     source = False
 
-    def __init__(self, code, short_desc, context, *parameters):
+    def __init__(
+            self,
+            code: str,
+            short_desc: str,
+            context: str,
+            *parameters: Iterable[str]
+    ) -> None:
         """Initialize the object.
 
         `parameters` are specific to the created error.
@@ -29,10 +38,10 @@ class Error:
         self.short_desc = short_desc
         self.context = context
         self.parameters = parameters
-        self.definition = None
-        self.explanation = None
+        self.definition = None  # type: Optional[Definition]
+        self.explanation = None  # type: Optional[str]
 
-    def set_context(self, definition, explanation):
+    def set_context(self, definition: Definition, explanation: str) -> None:
         """Set the source code context for this error."""
         self.definition = definition
         self.explanation = explanation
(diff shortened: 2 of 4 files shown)


requirements/tests.txt
```diff
@@ -4,3 +4,4 @@ mock==2.0.0
 pathlib
 mypy==0.630
 six
+typing
```
shacharoo: My bad 😋

src/pydocstyle/violations.py
```diff
@@ -68,9 +79,10 @@ def lines(self):
                 break
         return source
 
-    def __str__(self):
-        self.explanation = '\n'.join(l for l in self.explanation.split('\n')
-                                     if not is_blank(l))
+    def __str__(self) -> str:
+        if self.explanation:
+            self.explanation = '\n'.join(l for l in self.explanation.split('\n')
```
shacharoo: Unrelated to this PR, but shouldn't this only happen once?

src/pydocstyle/violations.py
```diff
@@ -29,10 +38,10 @@ def __init__(self, code, short_desc, context, *parameters):
         self.short_desc = short_desc
         self.context = context
         self.parameters = parameters
-        self.definition = None
-        self.explanation = None
+        self.definition = None  # type: Optional[Definition]
+        self.explanation = None  # type: Optional[str]
```
shacharoo: Can't we use the `self.explanation: Optional[str] = None` annotation?

src/pydocstyle/violations.py
```diff
@@ -82,10 +94,10 @@ def __str__(self):
                                ['filename', 'line', 'definition', 'message',
                                 'explanation', 'lines']})
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return str(self)
 
-    def __lt__(self, other):
+    def __lt__(self, other: 'Error') -> bool:
```
shacharoo: Why is Error a string?

src/pydocstyle/violations.py
```diff
@@ -19,7 +22,13 @@ class Error:
     explain = False
     source = False
 
-    def __init__(self, code, short_desc, context, *parameters):
+    def __init__(
+            self,
```
shacharoo: Isn't this over-indented?

src/pydocstyle/violations.py
```diff
@@ -107,9 +119,14 @@ def __init__(self, prefix, name):
             """
             self.prefix = prefix
             self.name = name
-            self.errors = []
-
-        def create_error(self, error_code, error_desc, error_context=None):
+            self.errors = []  # type: List[ErrorParams]
+
+        def create_error(
+                self,
```
shacharoo: Same as above (indentation)

src/pydocstyle/violations.py
Nurdok: Non-comment variable type hints were introduced only in Python 3.6 (with [PEP 526](https://www.python.org/dev/peps/pep-0526/)).

src/pydocstyle/violations.py
Nurdok: image

src/pydocstyle/violations.py
Nurdok: In a different PR :)

src/pydocstyle/violations.py
Nurdok: I think it should be indented once more than the code block to avoid confusion - this is PyCharm's default. Do you think otherwise?

src/pydocstyle/violations.py
Nurdok: Same as above

PR #349, opened 2018-12-19
Diff option

kuter:
Report changes only within line number ranges.

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -13,7 +13,7 @@ from .config import IllegalConfiguration
 from .parser import (Package, Module, Class, NestedClass, Definition, AllError,
                      Method, Function, NestedFunction, Parser, StringIO,
                      ParseError)
-from .utils import log, is_blank, pairwise, parse_unified_diff
+from .utils import log, is_blank, pairwise
 from .wordlists import IMPERATIVE_VERBS, IMPERATIVE_BLACKLIST, stem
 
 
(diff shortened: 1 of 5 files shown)


sigmavirus24:
Why not use flake8 + flake8-docstrings instead of copying this code around. Alternatively, @Nurdok @shacharoo should we pull this into a library so pydocstyle could use it.

Nurdok:
@kuter first of all, thanks for putting time and effort into this.

However, I think I agree with @sigmavirus24 - at the very least I would like the generic parsing utilities to be imported from a separate library. But I'm not sure reporting on a range on lines is even worth doing - this will cause files to be inconsistent within themselves. Is it worth it?

Can you somehow demonstrate this feature has enough audience of interested users?

kuter:
@Nurdok flake8 and pycodestyle has that option. I think that feature might be useful for people who works with old, undocumented code .. it's a time-consuming to add docstring to every undocumented method/function/class in huge file (let's say over 500 lines). Diff feature allows to check only changes before pushing them to repo like:

`hg diff -r `hg branch`:default | pydocstyle --diff`

sigmavirus24:
@kuter so it sounds like you're already using flake8. flake8-docstrings will do what you want without adding technical burden to pydocstyle's maintenance and will report only the errors in the diff. If you're using flake8 today, you could do:

```
hg diff -r current_branch;default | flake8 --diff
```

And get Docstring and normal violations reported in a single command.

@Nurdok I've used this on projects where there are docstrings but they violate pydocstyle and want to ensure new code doesn't violate that while allowing the old docstrings to co-exist for the time being.

kuter:
@sigmavirus24 I didn't know about flake8-docstring but seems like it meet my needs. Thanks.

Nurdok:
@kuter I'm glad you found a solution. I'm closing this for now.

thisch:
I think it would be nice if this PR could be reopened, because otherwise people that are using `pylint` would have to run both `flake8` as well as `pylint`.

PR #353, opened 2019-03-11
Invalid escape sequence fixes

scop:
Thanks for submitting a PR!

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -11,6 +11,7 @@ Major Updates
 
 * Support for Python 2.x and PyPy has been dropped (#340).
 * Added pre-commit hook (#346)
+* Invalid escape sequence fixes (#353)
 
 3.0.0 - October 14th, 2018
 --------------------------
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -428,7 +428,7 @@ class ConventionChecker:
 
         For example, if `line` is "  Hello world!!!", returns "Hello world".
         """
-        result = re("[\w ]+").match(line.strip())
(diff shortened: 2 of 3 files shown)


scop:
Done, hope I got it right now. Let me know if you'd prefer a clean rebase instead of what's here now.

Nurdok:
Looks correct, but what issue does this fix exactly? Can you add some tests?

scop:
Avoids warnings:
```
$ python3 -Wd -c 'foo = "[^\.].*"'
<string>:1: DeprecationWarning: invalid escape sequence \.
```
...which will become errors at some point, https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
> Changed in version 3.6: Unrecognized escape sequences produce a DeprecationWarning. In some future version of Python they will be a SyntaxError.

Wrt testing, I think it's indirect and kind of out of scope for this PR, but running tests with warnings enabled and someone looking at the logs would do somewhat. Or maybe turn them to errors. Probably requires upgrading pytest to >= 3.1 and... something, maybe a -Wd flag to pytest (although that didn't catch these for some reason in a quick test).

asottile:
In python3.8 this becomes a SyntaxWarning and in future versions this will be a SyntaxError

samj1912:
Re: tests, now that we have py3.8 in the CI, should we just add a -W flag to pytest so that it fails on any warnings that are emitted? This will ensure that there are no future regex additions like these.

samj1912 (approved):

samj1912:

asottile:
it looks like #445 superseded this

PR #355, opened 2019-03-15
fix parsing tuple syntax dunder all

### LINKED ISSUE #352: __all__ is parsed incorrectly
Causes warning:
```python
__all__ = 'exportable1', 'exportable2'
```

Works well:
```python
__all__ = ('exportable1', 'exportable2')
```

But both use valid tuple syntax so in the first case, `pydocstyle` is not supposed to yell at user.

Hanaasagi:
Resolves #352

### DIFF
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -444,26 +444,35 @@ class Parser:
             self.dunder_all_error = 'Could not evaluate contents of __all__. '
             return
         self.consume(tk.OP)
-        if self.current.value not in '([':
-            self.dunder_all_error = 'Could not evaluate contents of __all__. '
-            return
-        self.consume(tk.OP)
+
+        is_surrounded = False
+        if self.current.value in '([':
+            is_surrounded = True
+            self.consume(tk.OP)
 
         dunder_all_content = "("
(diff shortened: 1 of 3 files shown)


src/tests/parser_test.py
```diff
@@ -538,6 +559,29 @@ def test_dunder_all(code):
     assert module.dunder_all == ('foo', 'bar')
 
 
+def test_dunder_all_without_parentheses():
```
Nurdok: I'm not sure `without_parenthesis` describes what this is testing. Maybe `single_value`?

src/pydocstyle/parser.py
```diff
@@ -472,7 +481,7 @@ def parse_dunder_all(self):
                 '\bThe value was {}. The exception was:\n{}'
                     .format(dunder_all_content, e))
 
-        while not self.current.kind in self.stream.LOGICAL_NEWLINES:
+        while self.current.kind not in self.stream.LOGICAL_NEWLINES and self.current.kind != tk.ENDMARKER:
```
Nurdok: Please keep to 80 characters per line.

Nurdok:
Merged!

@Hanaasagi, thanks for your time and effort!

PR #357, opened 2019-04-19
Add an initial Google convention

samj1912:
### Sources for docstring convention:
[Google Style Guide](http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) and [Napoleon Google Style Guide](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

### Rationale for exclusions of errors:
 - **D203:** Multiple examples with class docstrings without blank line.
 - **D204:** Same as above.
 - **D213:** All multiline summaries start immediately after the triple quotes.
 - **D215:** N/A no underlines.
 - **D400:** First line can end with "period, question mark, or exclamation
point".
 - **D401:** Style guide explicitly says not to use imperative mood and use
descriptive mood.
 - **D404:** Clearly allowed WRT to the above. There is also an example using
"This".
 - **D406:** Section names end with a colon.
 - **D407:** N/A no underlines.
 - **D408:** N/A no underlines.
 - **D409:** N/A no underlines.

### Adds the following violations:
 * **D415:** First line of docstring should end with a period, question mark, or exclamation point 
 * **D416:** Section name should end with a semicolon
 * **D417:** Missing arguments in the function docstring

Thanks for submitting a PR!


samj1912:
TODO for initial PR:

 - [ ] Add google style args check
 - [ ] Add google style section check

samj1912:
While working on extension, I am beginning to realize that treating conventions as a set of exclusions may not be the best way to go. Do any of the core devs have any objections to passing the convention name and having conditional violations being raised only for a certain type of convention?

samj1912:
While writing this patch I am coming to the realization that conventions as they are currently defined in pydocstyle might lead to too complex code. Rather than having them as a list of errors to ignore it makes more sense for each of them to define a certain set of errors instead, following which the checkers should be segregated to match the conventions.

But this **might** be a backwards incompatible change especially since `ignores` and `conventions` are mutually exclusive.

Can any of the maintainers of the repo chip-in here with the best way forward?

I am meanwhile continuing to write the checkers given the current constraints of conventions and ignores, but it will obviously be easy to refactor it once a decision on the above is made.

sigmavirus24:
I'm not a maintainer, but I'm wondering if style conventions might be better if they were treated like plugins such that new styles (like this one) can be distributed separately from pydocstyle

samj1912:
I am not sure a plugin system exists for pydocstyle, but yeah if it did this PR would've been better as a plugin.

sigmavirus24:
> I am not sure a plugin system exists for pydocstyle, 

I took a quick look and I think the answer is no

samj1912:
I have added google style sections. I guess this PR is big enough as it is. For further section level checks 
(like paramter checks in Args/Attributes sections etc.) I will be creating a separate PR.

samj1912:
Once this is reviewed, I plan on submitting another PR refactoring the ConventionChecker into more modular parts and maybe we can discuss how to handle these styles in general.

But this PR should be good to go as it is.

src/pydocstyle/checker.py
```diff
@@ -633,29 +687,84 @@ def _suspected_as_section(_line):
...
+        """Check for docstring sections."""
+        if not docstring:
+            return
+
+        lines = docstring.split("\n")
+        if len(lines) < 2:
+            return
+
+        yield from self._check_numpy_sections(lines, definition, docstring)
+        yield from self._check_google_sections(lines, definition, docstring)
```
samj1912: Using `yield from` since docs say we will drop python2.7 support from the next release.

shacharoo:
@sigmavirus24 I think Pydocstyle shouldn't become a plug-and-play system for running convention checkers. Code should comply with the conventions we check in it, not manipulate it to fit their specific needs.

@samj1912 - I'll try to make some time this weekend to take a look at it. Thanks for contributing!

samj1912:
@shacharoo thanks :) You will find that all the commits are self-contained. So in case the PR is too big to review, I can chop the PR up by commits easily.

sigmavirus24:
@shacharoo I'm not advocating auto-fixing, I'm advocating the --convention flag become extendable

src/pydocstyle/checker.py
```diff
@@ -38,19 +38,57 @@ class ConventionChecker:
...
-                     'Yields',
-                     'Other Parameters',
-                     'Raises',
-                     'See Also',
-                     'Notes',
-                     'References',
-                     'Examples',
-                     'Attributes',
-                     'Methods']
+    NUMPY_SECTION_NAMES = [
```
Nurdok: This is immutable, so it can be a tuple.

src/pydocstyle/checker.py
```diff
@@ -38,19 +38,57 @@ class ConventionChecker:
...
+        'Tip',
+        'Todo',
+        'Warning',
+        'Warnings',
+        'Warns',
+        'Yield',
+        'Yields',
+    ]
+
+    GOOGLE_SECTION_REGEX = re(r'^(\s|\w)+:\s*$')
```
Nurdok: Please add an explanation comment for each regular expression with an example.

src/pydocstyle/checker.py
```diff
@@ -335,17 +373,31 @@ def check_unicode_docstring(self, definition, docstring):
                     ('u', 'ur')):
                 return violations.D302()
 
+    @staticmethod
+    def _check_ends_with(docstring, chars, violation):
```
Nurdok: Add a docstring please.

src/pydocstyle/checker.py
```diff
@@ -583,41 +618,78 @@ def _check_section(cls, docstring, definition, context):
         if not is_blank(context.previous_line):
             yield violations.D411(capitalized_section)
 
+    @classmethod
+    def _check_numpy_section(cls, docstring, definition, context):
+        """D4{05,06,10,11,13}, D214: Section name checks.
+
+        Check for valid section names. Checks that:
+            * The section name is properly capitalized (D405).
```
Nurdok: Please don't repeat all the description from `_check_common_section` as it's likely to go out of date. You can add to "Also yield all the errors from ..." line below.

src/pydocstyle/checker.py
```diff
@@ -583,41 +618,78 @@ def _check_section(cls, docstring, definition, context):
...
-        """D21{4,5}, D4{05,06,07,08,09,10}: Docstring sections checks.
-
-        Check the general format of a sectioned docstring:
-            '''This is my one-liner.
-
-            Short Summary
-            -------------
-            This is my summary.
+    @staticmethod
+    def _parse_args_section(docstring, definition, context):
```
Nurdok: Please add a docstring.

src/tests/test_cases/test.py
```diff
@@ -368,13 +374,13 @@ def docstring_bad_ignore_all():  # noqa
...
 
-def docstring_bad_ignore_one():  # noqa: D400,D401
+def docstring_bad_ignore_one():  # noqa: D400,D401,D415
     """Runs something"""
     pass
 
 
 @expect("D401: First line should be in imperative mood ('Run', not 'Runs')")
-def docstring_ignore_violations_of_pydocstyle_D400_and_PEP8_E501_but_catch_D401():  # noqa: E501,D400
+def docstring_ignore_violations_of_pydocstyle_D400_D415_and_PEP8_E501_but_catch_D401():  # noqa: E501,D400,D415
```
Nurdok: This is getting out of hand :)
Maybe change to `ignore_some_violations_but_catch_D401`?

Nurdok:
> I'm not advocating auto-fixing, I'm advocating the --convention flag become extendable

Why? I think PEP-257, numpy and the Google convention are popular enough and have a big enough traction in their own fields to merit enforcement. But I definitely **don't** want to encourage folks to create their own conventions. If I could eliminate all but one convention, I would :)

Pydocstyle has the option to enable and disable violations to support a gradual ramp up to full conformance and I accept that some codebases will never fully conform (and *some* conformance may be better than none). But I'm on the side of having an opinionated linter - that only helps you if you're trying to stick to a known convention. If some developers somewhere want to support a new convention other than the ones mentioned here, I sincerely hope they fail and adopt PEP-257 instead 😄

sigmavirus24:
>  If I could eliminate all but one convention, I would :)

What if I told you, by allowing plug-able conventions, you could. ;)

> If some developers somewhere want to support a new convention other than the ones mentioned here, I sincerely hope they fail and adopt PEP-257 instead

There are code-bases much older than PEP-257 and older than this tool. Those code-bases probably have conventions they'd like to enforce but have to do manually in a completely error-prone way and updating them is toilsome and a potentially ridiculous amount of work. I, for one, really don't think that they should have to fork this tool in order to gain the benefits of automated docstring style enforcement.

Only having 1 (or 2, or 3, ... or 5) standards works well enough when it's considered very early on (see also golint, rust-lang's doc format, etc.) but for Python we don't have that luxury and saying "I'll only support these *hand-waving* number of conventions because they're the most popular" comes across really poorly to the rest of the Python world.

samj1912:
FWIW I implemented a plugin system in https://github.com/PyCQA/pydocstyle/commit/c4aa001bac8420550d33054205b080951c1d86c7

The PR is at https://github.com/PyCQA/pydocstyle/pull/359 (although it says 800 loc most of the code is from #358 ) and the actual plugin system after the refactor is 30 loc.

samj1912:
@Nurdok @shacharoo all existing comments have been resolved.

samj1912:
Also, I added some missing support for D414 to both numpy and google style sections.

samj1912:
@Nurdok @shacharoo ping.

samj1912:
@Nurdok @shacharoo ping.

src/pydocstyle/checker.py
```diff
@@ -29,7 +29,7 @@ def decorator(f):
 
 
 class ConventionChecker:
-    """Checker for PEP 257 and numpy conventions.
+    """Checker for PEP 257, numpy and google conventions.
```
Nurdok: Capitalize Google? Can also change to NumPy I guess.

src/pydocstyle/checker.py
```diff
@@ -38,19 +38,71 @@ class ConventionChecker:
...
+        'Tip',
+        'Todo',
+        'Warning',
+        'Warnings',
+        'Warns',
+        'Yield',
+        'Yields',
+    )
+
+    # Matches anything that fulfills all the following conditions:
```
Nurdok: These docs are great. I think these can be formatted better by interleaving these with the actual regex. Something like:

```.py
GOOGLE_ARGS_REGEX = re(
  r'^\s'  # Begins with...
  r'(\w)'  # Followed by...
  ...
)
```

This'll save you from copying the regex to the comments and help the docs stay relevant in case this changes.

samj1912:
@Nurdok done :)

samj1912:
@Nurdok @shacharoo any chance you can take a look at this, this weekend?

Nurdok:
@samj1912 Thank you so much for doing this. Really sorry for the long delays (@shacharoo just got married and I was abroad several times, to our defense 😄). Merging.

mkhorton:
Just a quick note that your [usage doc](http://www.pydocstyle.org/en/4.0.1/usage.html) hasn't been updated with this new option, found this PR by accident. Thanks for adding support for google conventions!

PR #358, opened 2019-04-22
Refactor the checker

samj1912:
Does the following for the checkers -
 * Improves re compile. Currently we were doing a re compile inside
function calls. Moves this out so that this compile only happens once.
 * Improve checker code quality -
   * Current checker is stateless. It was still written as a class which
	added additional complexity with addition of passing of `self`
	`cls` and using staticmethod decorators.
   * Move the checker hook out of the class.
   * Slice up the checks into base, other and numpy style checks.
 * Use `find_packages` for aut-discovery of packages.
 * Switch to absolute imports.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,5 @@
 from __future__ import with_statement
-from setuptools import setup
+from setuptools import setup, find_packages
 import sys
 
 # Do not update the version manually - it is managed by `bumpversion`.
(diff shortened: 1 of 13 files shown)


samj1912:
While writing the google style checks PR, I felt that the ConventionChecker although a class doesn't seem to use any state at all and is instead used as a code organization unit. This leads to added complexity with @staticmethods and @classmethods as described above and passing of unused `self` and `cls` variables.

Given that all the code seemed functional, I took liberty into moving them into modules (which as intended as code organization pieces in python).

This also divides the checks into certain "styles" -
 * base - Base pep257 checks and common checks go here.
 * numpy - numpy only checks go here
 * other - One of checks used by neither go here.

samj1912:
@shacharoo @Nurdok comments appreciated. Rationale for the refactor is described above. 
I have currently kept ConventionChecker as a shell to maintain backwards compatibility in case any of the other tools are using it (which from a quick search they are.) I plan on adding deprecation warnings to them.

samj1912:

PR #374, opened 2019-07-07
Set `python_requires` so pip will not consider this on unsupported pythons

asottile:


asottile:
since a source distribution was not uploaded for 4.0.0 this can still help users if one were uploaded in the future

I suspect you're going to get requests from operating system packagers to upload a source distribution anyway ;)

Nurdok:
Thanks, @asottile!
Merging.

PR #381, opened 2019-07-17
Add error D303 when an f-string is used as a docstring

### LINKED ISSUE #368: ValueError: malformed node or string
Hi! I am running `pydocstyle` as a part of `flake8-docstrings` plugin which is a plugin for `wemake-python-styleguide`. 

Original issue: https://github.com/wemake-services/wemake-python-styleguide/issues/579

One of my users reported this bug:

```
multiprocessing.pool.RemoteTraceback:
"""
File "/usr/lib/python3.7/multiprocessing/pool.py", line 121, in worker
result = (True, func(*args, **kwds))
File "/usr/lib/python3.7/multiprocessing/pool.py", line 44, in mapstar
return list(map(*args))
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/checker.py", line 669, in _run_checks
return checker.run_checks()
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/checker.py", line 608, in run_checks
self.run_ast_checks()
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/checker.py", line 504, in run_ast_checks
for (line_number, offset, text, check) in runner:
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8_docstrings.py", line 89, in run
for error in self._check_source():
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8_docstrings.py", line 78, in _check_source
ignore_decorators=None,
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/pydocstyle/checker.py", line 75, in check_source
definition.docstring)
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/pydocstyle/checker.py", line 114, in check_docstring_missing
docstring and is_blank(ast.literal_eval(docstring))):
File "/usr/lib/python3.7/ast.py", line 91, in literal_eval
return _convert(node_or_string)
File "/usr/lib/python3.7/ast.py", line 90, in _convert
return _convert_signed_num(node)
File "/usr/lib/python3.7/ast.py", line 63, in _convert_signed_num
return _convert_num(node)
File "/usr/lib/python3.7/ast.py", line 55, in _convert_num
raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: <_ast.JoinedStr object at 0x7f1d8db27358>
"""

The above exception was the direct cause of the following exception:

File "/home/ilya/.virtualenvs/ds-backend/bin/flake8", line 10, in 
sys.exit(main())
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/main/cli.py", line 18, in main
app.run(argv)
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/main/application.py", line 394, in run
self._run(argv)
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/main/application.py", line 382, in _run
self.run_checks()
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/main/application.py", line 301, in run_checks
self.file_checker_manager.run()
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/checker.py", line 328, in run
self.run_parallel()
File "/home/ilya/.virtualenvs/ds-backend/lib/python3.7/site-packages/flake8/checker.py", line 292, in run_parallel
for ret in pool_map:
File "/usr/lib/python3.7/multiprocessing/pool.py", line 354, in 
return (item for chunk in result for item in chunk)
File "/usr/lib/python3.7/multiprocessing/pool.py", line 748, in next
raise value
ValueError: malformed node or string: <_ast.JoinedStr object at 0x7f1d8db27358>
```

This happens here: https://github.com/PyCQA/pydocstyle/blob/master/src/pydocstyle/checker.py#L159

Reproduction code:

```python
def function():
    f"""Template #{locals}"""
    ...
```

lordmauve:
Emit a new error code D303 when an f-string is found where a docstring is expected.

Fixes #368.


lordmauve:
Tests are green locally under 3.5-3.7 except for the stemming issue that #379 addresses.

src/pydocstyle/checker.py
```diff
@@ -28,6 +28,24 @@ def decorator(f):
     return decorator
 
 
+
+FSTRING_RE = re(r'^[rR]?[fF]')
```
sobolevn: Maybe case insensitive regex will be better?

sobolevn:
Thanks a lot, @lordmauve! I rely on this change in https://github.com/wemake-services/wemake-python-styleguide/issues/579

src/pydocstyle/checker.py
lordmauve: Case insensitive regexes invoke Unicode case conversion rules and could allow in other code points that may not be valid. Possibly `re.IGNORECASE | re.ASCII` might work but that's more verbose than this.

sobolevn:
@Nurdok friendly ping. What are the steps to get this merge?

Some of my users rely on this fix: https://github.com/wemake-services/wemake-python-styleguide/issues/579

lordmauve:
Grr, I suck at git, trying to get this rebased and mergable... I don't know where 14a59e6 came from.

lordmauve:
Ok, I think that's clean

lordmauve:
Tests are now failing because mypy returns an error when it sees fstrings under Python 3.5, even though the file won't be imported under Python 3.5, and it appears impossible to tell mypy to ignore this. Looking for a workaround.

sobolevn:
`# type: ignore` to the rescue!

lordmauve:
That does not work; mypy continues to parse the file and raise the error.

Nurdok:
Sorry for the long delays in response - I've been on back-to-back family vacation and business trips. I'll to catch up as soon as I can.

src/tests/test_cases/test.py
```diff
@@ -429,5 +429,5 @@ def bad_google_string():  # noqa: D400
     """Test a valid something"""
 
 
-expect(os.path.normcase(__file__ if __file__[-1] != 'c' else __file__[:-1]),
+expect(os.path.relpath(__file__),
```
Nurdok: Can you explain this change? This logic was handling a subtle bug where is subsequent execution, `__file__` contains the `.pyc` version which cannot be parsed by `pydocstyle`.

src/pydocstyle/checker.py
```diff
@@ -154,6 +183,9 @@ def check_docstring_missing(self, definition, docstring):
               with a single underscore.
 
         """
+        if is_fstring(docstring):
```
Nurdok: Why is this needed?

src/tests/test_definitions.py
```diff
@@ -20,17 +22,41 @@
...
+    run_case(test_case)
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 6),
+    reason="f-string support needed"
+)
+def test_fstrings():
+    """Run the f-string test case under Python 3.6+ only."""
+    # When run under Python 3.5, mypy reports a parse error for the test file,
```
Nurdok: Doesn't mypy have a file-ignore functionality? This seems overly complex. We can alternatively just avoid running mypy on Python < 3.6.

src/tests/test_cases/fstrings.py36
```diff
@@ -0,0 +1,11 @@
...
+
+from .expected import Expectation
+
+expectation = Expectation()
+expect = expectation.expect
+
+
+@expect("D303: f-strings are not valid as docstrings")
+def fstring():
+    f"""Toggle the gizmo."""
```
Nurdok: Test more variants of f-strings please.

src/pydocstyle/checker.py
```diff
@@ -29,6 +29,24 @@ def decorator(f):
...
 
 
+
+FSTRING_RE = re(r'^[rR]?[fF]')
+
+
+def is_fstring(docstring):
+    r"""Return True if docstring is an f-string.
+
+    >>> is_fstring('rf"abc"')
```
Nurdok: We don't actually run doctests IIRC.

docs/release_notes.rst
```diff
@@ -19,6 +19,8 @@ New Features
   docstrings (#407).
 * Added support for Python 3.8 (#423).
 * Allow skipping errors on module level docstring via #noqa (#427).
+* New error code D303 is emitted when an f-string is found in place of a
+  docstring.
```
Nurdok: Please add the issue and PR IDs here.

src/tests/test_definitions.py
Nurdok: What about this?

> A hacky way to make mypy ignore the module would be to add a stub file to the same directory, e.g. if you want mypy to ignore foo.py, add an empty foo.pyi.

https://github.com/python/mypy/issues/964

src/pydocstyle/checker.py
lordmauve: I think that is there to avoid getting two errors if people use an f-string as a docstring.

src/tests/test_definitions.py
lordmauve: I can try that and see if it works. If it doesn't, are you happy with this solution or should I think again?

lordmauve: Thinking again could involve `eval()` of a triple-quoted string. Which ends up being a separate, parallel test harness. I thought this was nicer for using one test harness with just this filename trick.

lordmauve: Not running mypy on <Python 3.6 means that mypy will not catch things like use of APIs or parameters that were added in 3.6. I assumed you would be unhappy with that.

samj1912:
@lordmauve any updates on this?

samj1912:

src/tests/test_definitions.py
samj1912: Since we have dropped support for py<3.6 I was able to remove all this logic.

samj1912 (approved):
LGTM and I have also addressed all of @Nurdok's comments

samj1912:
@Nurdok please take a look when you have time - I make the code more resilient to fstrings in case people ignore it even after the warnings.

samj1912:
@Nurdok will you have a chance to look at this? It LGTM but waiting on your review.

src/pydocstyle/checker.py
```diff
@@ -44,6 +44,27 @@ def decorator(f):
...
+
+def safe_literal_eval(string):
+    """Safely evaluate a literal even if it is an fstring."""
+    try:
+        return ast.literal_eval(string)
+    except ValueError as error:
+        # If the docstring is a fstring, it is
+        # not considered a valid docstring. See
+        # https://bugs.python.org/issue28739
+        raise ParseError(
```
Nurdok, line 63: Mistakenly using an f-string for a docstring does not cause a syntax error, so we shouldn't raise a `ParseError` here.
Instead, I suggest we rename this as `eval_docstring` and just return an empty string if we find an f-string.

src/pydocstyle/checker.py
```diff
@@ -176,8 +197,24 @@ def checks(self):
             for this_check in vars(type(self)).values()
             if hasattr(this_check, '_check_for')
         ]
+        # This returns the checks in the order they are
+        # listed in the file (since py3.6) if their priority is the same
         return sorted(all, key=lambda this_check: not this_check._terminal)
 
+    # Note - this needs to be listed before other checks
```
Nurdok, line 204: Checks should not rely on order, other checker functions, etc. Let's find a better solution.
Perhaps my suggestion for returning an empty string from `eval_docstring` is enough to solve this. Otherwise, we can have it return `None` and do more verbose checks at call sites. LMKWYT.

Nurdok (changes requested):

PR #383, opened 2019-07-21
Fix failing tests due to modifications to imperatives

samj1912:
With the recent addition of "initiate" to the list of imperatives,
both "initialize" and "initiate" stem to the same verb - "initi"
which causes an error since we use Initialize in our docs and due
to the way the imperative suggestion logic is implemented.

Since initiate occurs later in the file - when constructing a
dictionary - it trumps initialize and replaces it.

This patch fixes it by changing the dictionary to a list of verbs
instead of a single stem-verb map.

Thanks for submitting a PR!

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -440,16 +440,16 @@ class ConventionChecker:
                     return violations.D401b(first_word)
 
                 try:
-                    correct_form = IMPERATIVE_VERBS.get(stem(check_word))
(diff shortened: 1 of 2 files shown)


samj1912:
@Nurdok with the latest push to the master - the tests broke. Can you take a look at this PR?

Because of the master all the other PRs are failing as well.

samj1912:
NVM, didn't see there was an open PR #382 which does this in a much better way. Closing this.

PR #389, opened 2019-07-31
Flake8 plugin

### LINKED ISSUE #219: Provide `flake8` Plugin
I was thinking that we should provide our own plugin to `flake8` instead of `flake8-docstrings`. It would solve issues like #207 where we break `flake8-docstrings` by accident.

Also, there is an issue that bothers me with the current `flake8` plugin - it ignores `pydocstyle` configuration files. While other projects might use configuration files as just a handy way of typing less, in `pydocstyle` the configuration files allow for flexible styles over different directories in a project. I personally don't use `flake8` over `pydocstyle` at work because of this issue. 

There is an issue of how error selection might work in conjunction with `flake8`. If a `pydocstyle` config file specifies `ignore=D100`, would the plugin simply not yield D100 errors? What happens when `flake8` is run with `select=D100`? It would not have any effect. Another option is using the configuration files' selection _unless_ `flake8` is used with `pydocstyle` error selection. Yet another way is to treat `flake8`'s selections as `add-*` so that it amends the selection in the configuration files instead of replacing or masking them.

@sigmavirus24 Any thoughts? I'm currently leaning toward the last one - i.e., `pydocstyle` config file is read, then amended by `flake8` CLI (or config files).

chaoflow:
Thanks for submitting a PR!

close #219

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,5 +42,8 @@ setup(
         'console_scripts': [
             'pydocstyle = pydocstyle.cli:main',
         ],
+        'flake8.extension': [
+            'D = pydocstyle.flake8_plugin:Checker',
+        ],
     },
 )
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -101,7 +101,7 @@ class ConfigurationParser:
         options, _ = self._parse_args([])
         return self._create_run_config(options)
 
-    def parse(self):
+    def parse(self, args=None, values=None):
         """Parse the configuration.
 
         If one of `BASE_ERROR_SELECTION_OPTIONS` was selected, overrides all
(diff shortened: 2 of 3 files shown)


chaoflow:
Please see my [comment](https://github.com/PyCQA/pydocstyle/issues/219#issuecomment-517000505) on the issue. I'll fix tests and release notes, if this is generally welcome.

chaoflow:
The plugin in its current state works when called via flake8 on CLI and via ``pytest --flake8``. However, Emacs flycheck passes the file on stdin. So far I'm not aware of how to find out the filename / path. Documentation feels scarce.

Nurdok:

Issue #370, opened 2019-06-16
IndentationError: unexpected indent when docstring is in nested function.


pinkavaj:
Exception is raised for latest (6de6d93) pydocstyle when nested function contains docstring with arguments section.

## How to reproduce:

Write python file `sample.py`:
```
def foo():
    def bar(a):
        """A docstring

        Args:
            a : An argument.
        """
        pass
```

Run:  `pydocstyle sample.py`

## Result:

```
Traceback (most recent call last):
  File "/home/pinky/.local/bin/pydocstyle", line 11, in <module>
    load_entry_point('pydocstyle', 'console_scripts', 'pydocstyle')()
  File "/home/pinky/work/pydocstyle/src/pydocstyle/cli.py", line 68, in main
    sys.exit(run_pydocstyle())
  File "/home/pinky/work/pydocstyle/src/pydocstyle/cli.py", line 45, in run_pydocstyle
    ignore_decorators=ignore_decorators))
  File "/home/pinky/work/pydocstyle/src/pydocstyle/checker.py", line 883, in check
    ignore_decorators):
  File "/home/pinky/work/pydocstyle/src/pydocstyle/checker.py", line 124, in check_source
    for error in errors:
  File "/home/pinky/work/pydocstyle/src/pydocstyle/checker.py", line 831, in check_docstring_sections
    yield from self._check_google_sections(lines, definition, docstring)
  File "/home/pinky/work/pydocstyle/src/pydocstyle/checker.py", line 819, in _check_google_sections
    yield from self._check_google_section(docstring, definition, ctx)
  File "/home/pinky/work/pydocstyle/src/pydocstyle/checker.py", line 714, in _check_google_section
    yield from cls._check_args_section(docstring, definition, context)
  File "/home/pinky/work/pydocstyle/src/pydocstyle/checker.py", line 681, in _check_args_section
    function_pos_args = get_function_args(definition.source)
  File "/home/pinky/work/pydocstyle/src/pydocstyle/checker.py", line 916, in get_function_args
    function_arg_node = ast.parse(function_string).body[0].args
  File "/usr/lib/python3.7/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    def bar(a):
    ^
IndentationError: unexpected indent
```

Hanaasagi:
[...]

PR #392, opened 2019-08-02
Fix IndentationError while parsing function

sfstpala:
The `checker.get_function_args` function will fail with an IndentationError if it gets a `function_string` that is indented. This fixes that. :)

Thanks for submitting a PR!

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -13,6 +13,7 @@ Bug Fixes
   e.g., init and initialize / initiate (#382).
 * Fix parser hanging when there's a comment directly after ``__all__``
   (#391, #366).
+* Fixed IndentationError when parsing function arguments (#392).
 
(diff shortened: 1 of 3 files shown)


Nurdok:
Hi @sfstpala! Thanks for the PR. 

Can you please do the following?
1. Add a test (e.g., with the example from #370 or similar).
2. Rebase on the new master and resolve the conflicts.

sfstpala:
Thanks for taking a look @Nurdok - I added the test and resolved the conflict. Fixes #370

Hanaasagi:
@sfstpala  check out #390 . Seems you've duplicated my work ;-)

Nurdok:
Thanks, @sfstpala !

ljades:
Thank you so much for getting on this! When might we expect a version bump on PyPi with this fix?

Issue #361, opened 2019-04-29
Allow blank lines after multi-line function docstrings


anntzer:
Currently, pydocstyle warns by default when a function docstring is followed by a blank line (D202).  However, a careful reading of PEP257 indicates that this rule should only apply to one-line docstrings, not to multiline docstrings: the sole sentence regarding this, "There's no blank line either before or after the docstring.", occurs in the "One-line docstrings" section, immediately after the "The closing quotes are on the same line as the opening quotes. This looks better for one-liners." sentence which clearly only applies to one-line docstrings.  Allowing an empty line after the function docstring also makes sense if e.g. the function body contains empty lines: in something like
```
def func():
    """
    Some docstring.

    Some more stuff.
    """
    do_this()
    do_that()

    do_this_again()
    do_that_again()
```
if there's no empty line after the docstring, I personally feel the docstring "binds" too closely to the first part of the body at the expense of the second part.

cdeil:
In our project we are using `black` and `pydocstyle`.

There is the case of inner functions following docstrings where `black` and pydocstyle` disagree, one of the few problems that come up when using both tools.

Example: https://github.com/gammapy/gammapy/blob/a9ca5c02b191a0f03212163c106e35f45f05850c/gammapy/stats/significance.py#L111-L125

pydocstyle tells me to remove the empty line after the docstring:
```
gammapy/stats/significance.py:112 in public function `significance_to_probability_normal_limit`:
        D202: No blank lines allowed after function docstring (found 1)
```
And then I run black, it will add the empty line back, because it puts empty lines around inner functions:
https://github.com/python/black#empty-lines

To me, the `black` style seems reasonable, so I'm bringing this up here instead of with black.

Would you be willing to change D202 and make it so that this formatting is allowed and will not raise a D202?
```
def spam():
    """Spam."""

    def ham():
        return "ham"

    return "spam"
```

daviddavis:
@cdeil ran into the same issue. At the very least, pydocstyle should allow a blank line after docstring if there's a nested function.

ala-ableton:
AFAICS, this issue is fixed by https://github.com/PyCQA/pydocstyle/pull/426 and could be closed.

anntzer:
Actually I think this is not fully resolved: as noted in the original message, I think that a careful reading of PEP257 indicates that the prohibition for a blank link after the docstring only applies to single-line docstrings (the point literally appears in the "one-line docstrings" section).

[...]

PR #395, opened 2019-08-05
feat(checker): update D202 for inner functions

lewisacidic:
Changes D202: No blank lines allowed after function docstring to allow
space below function docstrings with inner functions:

i.e. allows

```python

def outer():
    """Valid docstring."""

    def inner():
        pass

    return pass
```

See comment from @cdeil in #361.

Thanks for submitting a PR!

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -199,7 +199,9 @@ class ConventionChecker:
             if blanks_before_count != 0:
                 yield violations.D201(blanks_before_count)
             if not all(blanks_after) and blanks_after_count != 0:
-                yield violations.D202(blanks_after_count)
+                def_follows = after.split("\n")[2].lstrip().startswith('def')
+                if not (blanks_after_count == 1 and def_follows):
+                    yield violations.D202(blanks_after_count)
 
     @check_for(Class)
     def check_blank_before_after_class(self, class_, docstring):
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -13,6 +13,8 @@ Bug Fixes
   e.g., init and initialize / initiate (#382).
 * Fix parser hanging when there's a comment directly after ``__all__``
   (#391, #366).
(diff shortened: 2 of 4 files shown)


lewisacidic:
Dirty-looking change to fix the compatibility clash with `black` when it comes to inner functions.

I didn't take long to look through the library for a nicer way to do this, just thought I would get the ball rolling. Would be nice to have, as I have to disable D202 for our projects otherwise.

src/pydocstyle/checker.py
```diff
@@ -199,7 +199,9 @@ def check_no_blank_before(self, function, docstring):  # def
             if blanks_before_count != 0:
                 yield violations.D201(blanks_before_count)
             if not all(blanks_after) and blanks_after_count != 0:
-                yield violations.D202(blanks_after_count)
+                def_follows = after.split("\n")[2].lstrip().startswith('def')
```
Code0x58: To allow both classes and functions you could use, and avoid including variables that have a name starting with _def_:
```python
re(r"\s+(?:class|def)\s").match(after)
```
This follows how python's re module is used in this file, where `re` is an alias for `re.compile`. It would be good to add a test for the `class` case too.

The check docstring could use updating to reflect the allowance of a single line if it is before a class/function definition.

johanfleury:
@lewisacidic are you still working on this?

I see there's not much left to do (fix the docstring at the top of `functions.py`, use `re` as asked by @Code0x58), I can probably take it over if you can't (or don't want to) work on this PR anymore?

lewisacidic:
Oh sorry completely forgot about this - I'm really busy until end of next week. I could come back to it then, but feel free to finish off if you want!

btw sorry for brutalizing this with force pushes...

johanfleury:
I create PR #426 with @Code0x58 recommendation.

peterjc:
Can this be closed with https://github.com/PyCQA/pydocstyle/pull/426 merged instead?

PR #396, opened 2019-08-12
Fix error table RST, max short code now over 69 chars

### LINKED ISSUE #380: Error codes missing from ReadTheDocs as of 4.0.0
On ReadTheDocs, [the current error codes page](http://www.pydocstyle.org/en/4.0.0/error_codes.html) has a blank **Grouping** heading.

For the [3.0.0 release docs](http://www.pydocstyle.org/en/3.0.0/error_codes.html), this is where the table of error codes would be printed.

peterjc:
Should close #380.

Thanks for submitting a PR!

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -13,6 +13,8 @@ Bug Fixes
   e.g., init and initialize / initiate (#382).
 * Fix parser hanging when there's a comment directly after ``__all__``
   (#391, #366).
+* Fixed RST error in table which resulted in the online documentation missing
+  the violation code table (#396).
 
 4.0.0 - July 6th, 2019
 ---------------------------
--- a/src/pydocstyle/violations.py
+++ b/src/pydocstyle/violations.py
@@ -152,18 +152,19 @@ class ErrorRegistry:
     @classmethod
     def to_rst(cls) -> str:
         """Output the registry as reStructuredText, for documentation."""
-        sep_line = '+' + 6 * '-' + '+' + '-' * 71 + '+\n'
-        blank_line = '|' + 78 * ' ' + '|\n'
+        max_len = max(len(error.short_desc) for group in cls.groups for  error in group.errors)
+        sep_line = '+' + 6 * '-' + '+' + '-' * (max_len + 2) + '+\n'
+        blank_line = '|' + (max_len + 9) * ' ' + '|\n'
         table = ''
         for group in cls.groups:
             table += sep_line
             table += blank_line
-            table += '|' + '**{}**'.format(group.name).center(78) + '|\n'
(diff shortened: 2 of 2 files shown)


peterjc:
If adding a dependency on https://pypi.org/project/restructuredtext-lint/ was acceptable, I could add a test for this?

lordmauve:
There are different requirements files for testing and documentation from the package itself, so I would guess that adding a dependency to one of those would not be a problem: https://github.com/PyCQA/pydocstyle/tree/master/requirements

peterjc:
Adding this to the tox ``docs`` jobs makes sense to me, but looking at all the RST files, we might need to tweak ``rst-lint`` a bit first to check all the files: https://github.com/twolfson/restructuredtext-lint/issues/52

```
$ python -c "from pydocstyle.violations import ErrorRegistry; print(ErrorRegistry.to_rst())" > docs/snippets/error_code_table.rst
$ rst-lint docs/*.rst docs/*/*.rst
ERROR docs/error_codes.rst:23 Unknown interpreted text role "ref".
ERROR docs/index.rst:19 Unknown directive type "toctree".
ERROR docs/snippets/config.rst:39 Unknown interpreted text role "ref".
ERROR docs/snippets/pre_commit.rst:232 Undefined substitution referenced: "version".
ERROR docs/snippets/config.rst:39 Unknown interpreted text role "ref".
ERROR docs/snippets/pre_commit.rst:6 Undefined substitution referenced: "version".
```

Installing restructuredtext-lint just for a couple of files seems rather heavy handed.

Perhaps running sphinx in a stricter mode would be better?

lordmauve:
> Perhaps running sphinx in a stricter mode would be better?

Yes - Sphinx's flavour of reStructuredText won't be understandable to plain docutils. I think Sphinx normally has a "dummy" build target that doesn't generate any output, but validates the input, but I don't see it in [pydocstyle's docs Makefile](https://github.com/PyCQA/pydocstyle/blob/master/docs/Makefile).

peterjc:
It seems better as a separate pull request, but maybe:

```
$ sphinx-build -h
...
  -n                nit-picky mode, warn about all missing references
...
  -W                turn warnings into errors
```

sigmavirus24:
Right, nothing keeps the Makefile up-to-date and users can modify it. This is why Flake8 uses `sphinx-build` directly in its `tox.ini` to validate what it cares about in the docs. Might be helpful to crib from there.

Nurdok:
I'll merge this in a few minutes when the build completes successfully (had to resolve release notes conflict).

@peterjc - I'm in favor of adding a linter to check for these kinds of errors in the future and don't mind depending on another package (for development / testing only, this is a no brainer). Feel free to send a followup PR for that.

Thanks!

Issue #177, opened 2016-03-02
tests break in Python 3 environment


danstender:
Hi, `test_integration` breaks in a Python 3 environment:
```
--- a/src/tests/test_integration.py
+++ b/src/tests/test_integration.py
@@ -100,7 +100,7 @@
 @pytest.yield_fixture(scope="module")
 def install_package(request):
     cwd = os.path.join(os.path.dirname(__file__), '..', '..')
-    install_cmd = "python setup.py develop"
+    install_cmd = "python3 setup.py develop"
     uninstall_cmd = install_cmd + ' --uninstall'
     subprocess.check_call(shlex.split(install_cmd), cwd=cwd)
     yield
```
BTW a new Debian package of pydocstyle is coming up.

sigmavirus24:
I suspect this is because the tests are run inside of tox for us. That said, I'm not sure this patch is correct for the project. I understand how Debian is building the packages is a bit naive when considering Python 2 versus Python 3, but doing this pytest fixture *should* work on Python 2 and Python 3. This patch would break the Python 2 tests.

Nurdok:
@sigmavirus24 is correct in that tests are supposed to run under `tox`, which will define `python` to be the correct Python version. However, I think a simple solution that will be applicable to all cases would be to use `sys.executable`.

danstender:
Yes, o.k., I'll take a second look if that patch is really needed to run/build on Py3. Wasn't mend to
adopt, either, but just as a pointer.

Thanks,
DS

Nurdok:
Out of curiosity, why is it needed to maintain Debian packages for Python packages? I develop on Ubuntu (sometimes), but I will always opt for using `pip` when I can.

danstender:
It's not pointless ...

Nurdok:
I'm not suggesting it is, I'm genuinely asking - when and why would you use a Debian package of a Python library over installing it via `pip`?

sigmavirus24:
@Nurdok lots of people trust signed packages over packages delivered over TLS from untrusted sources. PyPI is uncurated while Debian's repositories are ... somewhat curated. The repositories often check for namespace collisions with respect to command-line names (console scripts), library/module names, etc.

[...]

PR #403, opened 2019-09-15
test_integration: Use sys.executable

jayvdb:
Literal `python` is often `python2` and may not be present at all.

Closes https://github.com/PyCQA/pydocstyle/issues/177

### DIFF
--- a/src/tests/test_integration.py
+++ b/src/tests/test_integration.py
@@ -80,9 +80,10 @@ class SandboxEnv:
         run_target = self.tempdir if target is None else \
             os.path.join(self.tempdir, target)
 
-        cmd = shlex.split("{} {} {}"
-                          .format(self.script_name, run_target, args),
-                          posix=False)
+        cmd = shlex.split(
+            ('"{}" -m {} {} {}' if sys.platform == "nt" else '{} -m {} {} {}')
+            .format(sys.executable, self.script_name, run_target, args),
+            posix=False)
         p = subprocess.Popen(cmd,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


jayvdb:
Im not sure this warrants a relnote.

jayvdb:
(Working on a fix for Windows)

src/tests/test_integration.py
```diff
@@ -80,9 +80,10 @@ def invoke(self, args="", target=None):
         run_target = self.tempdir if target is None else \
             os.path.join(self.tempdir, target)
 
-        cmd = shlex.split("{} {} {}"
-                          .format(self.script_name, run_target, args),
-                          posix=False)
+        cmd = shlex.split(
+            ('"{}" -m {} {} {}' if sys.platform == "nt" else '{} -m {} {} {}')
```
Nurdok: Actually, this line was also testing that `pydocstyle` is installed as an executable by itself - there's no reason to use `sys.executable` here.

The change below in `test_run_as_named_module` is great though.

jayvdb: As this is a reusable component, can it be using the more reliable and faster CLI entry point `__main__` ?  This then covers the same paths as `test_run_as_named_module`

And then instead of `test_run_as_named_module` I create some dedicated tests for the generated stub created to invoke the entry point.

Then distro packagers can disable the tests for the generated stub, because e.g. atm openSUSE only creates the stub entry point for python3 , and therefore the py2 integration tests fail.

Nurdok: I don't mind making `invoke` use the `- m` way and changing `test_run_as_named_module` to, e.g., `test_run_as_script` if that's what you're suggesting. As long as they are both being tested.

jayvdb: Yup; that is what I had in mind.  I'll revise the PR over the w/e to that effect.

Nurdok: Ping, still working on this?

samj1912:

PR #404, opened 2019-09-19
Remove D413 from the pep257 convention

### LINKED ISSUE #401: D413 problems since pydocstyle 4.0.0
Updating `pydocstyle` from version 3.0.0 to version 4.0.0 or 4.0.1 throw numerous new `D413` (and few `D412`) issues on code that passes with version 3.0.0.

Example: The following code:
```python
"""Test module to demonstrate D413 problem."""


def example_1(one, two):
    """Demonstrate case.

    Arguments:
     - one      first argument
     - second   second argument

    This function shows if a D413 is thrown or not.
    In my view, this is not a D413.
    """
    pass


def example_2(one, two):
    """Demonstrate case.

    Attributes:
     - one      first attribute
     - second   second attribute

    This function shows if a D413 is thrown or not.
    In my view, this is not a D413.
    """
    pass


# Original PEP-257 example from https://www.python.org/dev/peps/pep-0257/:
def complex(real=0.0, imag=0.0):
    """Form a complex number.

    Keyword arguments:
    real -- the real part (default 0.0)
    imag -- the imaginary part (default 0.0)
    """
    if imag == 0.0 and real == 0.0:
        return complex_zero
```
throws 'only' one issue with `pydocstyle` ver. 3.0.0:
```bash
>pydocstyle --version
3.0.0

>pydocstyle D413_example.py
D413_example.py:18 in public function `example_2`:
        D413: Missing blank line after last section ('Attributes')
```

but **4** with ver. 4.0.1 (or 4.0.0), even with the original PEP-257 example (from https://www.python.org/dev/peps/pep-0257/):
```bash
>pydocstyle --version
4.0.1

>pydocstyle D413_example.py
D413_example.py:5 in public function `example_1`:
        D413: Missing blank line after last section ('Arguments')
D413_example.py:18 in public function `example_2`:
        D413: Missing blank line after last section ('Attributes')
D413_example.py:18 in public function `example_2`:
        D413: Missing blank line after last section ('Attributes')
D413_example.py:32 in public function `complex`:
        D413: Missing blank line after last section ('Keyword Arguments')
```

Actually, I think that **no** of these example docstrings should give an issue (at least in pep257 mode). The only way to get rid of this error is to put an empty line before the closing quotes, but this isn't a PEP 257 rule?

samj1912:
Closes #401

Thanks for submitting a PR!

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -16,6 +16,10 @@ New Features
 * Extend support for detecting missing arguments in Google style
   docstrings to method calls (#384).
 
+Bug Fixes
+
+* Remove D413 from the pep257 convention (#404).
+
 4.0.1 - August 14th, 2019
 -------------------------
 
--- a/src/pydocstyle/violations.py
+++ b/src/pydocstyle/violations.py
@@ -266,7 +266,7 @@ all_errors = set(ErrorRegistry.get_error_codes())
 conventions = AttrDict({
     'pep257': all_errors - {'D203', 'D212', 'D213', 'D214', 'D215', 'D404',
                             'D405', 'D406', 'D407', 'D408', 'D409', 'D410',
-                            'D411', 'D415', 'D416', 'D417'},
+                            'D411', 'D413', 'D415', 'D416', 'D417'},
     'numpy': all_errors - {'D107', 'D203', 'D212', 'D213', 'D402', 'D413',
                            'D415', 'D416', 'D417'},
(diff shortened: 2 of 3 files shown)


peterjc:
The second tick box is easy to satisfy now you have the pull request number.

One of the examples on #401 could be used for a test case perhaps? I'd use the PEP257 example deliberately as it shows why the (revised) D413 is not suitable for the PEP257 convention.

samj1912:
@Nurdok tests and release notes updated. Review pls.

Nurdok:
Thanks @samj1912!

PR #411, opened 2019-10-01
Fix decorator parsing

### LINKED ISSUE #410: Option ignore-decorators does not work properly
The ignore-decorators option does not work, when the decorated method is the first method in  a class.

Consider the following example file named foo.py:
```
"""Shut up."""


class FooClass:
    """Shut up."""

    @property
    def bar(self):
        pass

```
Then
```
$ pydocstyle --ignore-decorators='.*' foo.py
foo.py:8 in public method `bar`:
        D102: Missing docstring in public method
```

While the example foo2.py
```
"""Shut up."""


class FooClass:
    """Shut up."""

    def foo(self):
        """Shut up."""
        pass

    @property
    def bar(self):
        pass
```
is working as expected:
```$ pydocstyle --ignore-decorators='.*' foo2.py```

### PR DESCRIPTION
**Pull request by samj1912**
Fixes #410

The bug existed because we were not accounting for cases which should
be treated as a logical new line.

Thanks for submitting a PR!

### DIFF
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -267,7 +267,12 @@ class TokenStream:
         current = self._next_from_generator()
         self.current = None if current is None else Token(*current)
         self.line = self.current.start[0] if self.current else self.line
-        self.got_logical_newline = (previous.kind in self.LOGICAL_NEWLINES)
+        is_logical_blank = previous.kind in (tk.NL, tk.COMMENT)
+        self.got_logical_newline = (
+            previous.kind in self.LOGICAL_NEWLINES
+            # Retain logical_newline status if last line was logically blank
+            or (self.got_logical_newline and is_logical_blank)
+        )
         return previous
 
     def _next_from_generator(self):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by samj1912** [comment]
@Nurdok review please :)

File: src/pydocstyle/parser.py [hunk 1]
```diff
@@ -267,7 +267,27 @@ def move(self):
...
+        # v/s
+        # @decorator
+        # def test(): ...
+        self.got_logical_newline = (
+            previous.kind in self.LOGICAL_NEWLINES
+            # In case we had a logical new line previously and
+            # the previous token was a new line we should persist this state.
+            # This is so we don't skip parsing trivial decorator
+            # cases like as described
+            # in https://github.com/PyCQA/pydocstyle/issues/410
```
Nurdok: This can also be more concise, e.g.:

```
self.got_logical_newline = (
    previous.kind in self.LOGICAL_NEWLINES
    # Retain logical_newline status if last line was logically blank
    or (self.got_logical_newline and previous.kind in self.LOGICAL_BLANKS)
)
```

samj1912:
Addressed comments - ready for review again.

samj1912:
@Nurdok done

Issue #421, opened 2019-10-18
D401 does not properly check one-word documentations


Strus:
pydocstyle version: 4.0.1

D401 does not detect errors in documentation if it is just one word.

Example:

```python
"""Test file."""

class Foo:
    """Represents something."""

    def __init__(self, bar):
        """Constructor.

        :param bar: bar
        """
        self.bar = bar
```

```bash
pydocstyle --ignore D203,D213,D407,D416 test.py
<empty output>
```

When I add some more words to `__init__` documentation, it properly detects D401:

```python
"""Test file."""

class Foo:
    """Represents something."""

    def __init__(self, bar):
        """Constructor of Foo.

        :param bar: bar
        """
        self.bar = bar
```

```bash
pydocstyle --ignore D203,D213,D407,D416 test.py
test.py:7 in public method `__init__`:
        D401: First line should be in imperative mood; try rephrasing (found 'Constructor')
```

It also does not detect other single words like for example `Initalizer.`

PR #422, opened 2019-10-18
Fix D401 and D404 checks not working for docstrings containing only one word and ending with non-alpha character (#421)

Strus:


src/pydocstyle/utils.py
```diff
@@ -37,4 +39,10 @@ def common_prefix_length(a: str, b: str) -> int:
...
         if ca != cb:
             return common
-    return min(len(a), len(b))
\ No newline at end of file
+    return min(len(a), len(b))
+
+
+def strip_non_alphanumeric(string: str) -> str:
+    """Strip string from any non-alphanumeric characters."""
+    pattern = re.compile(r'[\W_]+')
```
samj1912: It doesn't make sense to do the inside the function definition. You won't benefit from doing a compile. I would move the compile outside, so that it's done only once.

Strus: Done. Sorry for lon response time, I was on vacation.

Nurdok:
To be honest, I'm not sure I like this change. The first line of a docstring should be a proper sentence that describes the function / class, and I don't see how one word, imperative of not, can qualify as a meaningful description. Happy to hear your thoughts on the subject. Perhaps we can add an issue that the first line should be > 1 word anyway?

Strus:
Without my changes if you add any non-alpha character after the first word, like comma, D401 and D404 will also not work properly, so my fix should be added anyway, even if you decide that one-word documentations should not be considered vaild.

Nurdok:
Good point, merging.

PR #425, opened 2019-10-23
Added simple fstring parsing testcase

yoseft:
Thanks for submitting a PR!

### DIFF
--- a/src/tests/parser_test.py
+++ b/src/tests/parser_test.py
@@ -49,8 +49,8 @@ def test_function():
     assert str(function) == 'in public function `do_something`'
 
 
-def test_fstring_docstring():
-    """Test parsing of a simple function with an fstring as its docstring."""
+def test_simple_fstring():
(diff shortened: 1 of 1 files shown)


src/tests/parser_test.py
```diff
@@ -49,6 +49,39 @@ def do_something(pos_param0, pos_param1, kw_param0="default"):
...
+    # fstrings are not supported in Python 3.5
+    if sys.version_info[0:2] == (3, 5):
+        return
+
+    parser = Parser()
+    code = CodeSnippet("""\
+        def do_something(pos_param0, pos_param1, kw_param0="default"):
+            f\"""Do something.\"""
+            return None
+    """)
```
Nurdok: Please also test this out with actual use of fstring capabilities, e.g.:

```py
foo = "bar"
def do():
    f"""Hello, {foo}"""
```

PR #426, opened 2019-10-24
fix(checker): allow whitespace before inner functions and class

johanfleury:
Changes D202: ("No blank lines allowed after function docstring") to
allow space below function docstrings with inner functions.

i.e. allows:

```python
def outer():
    """Valid docstring."""

    def inner():
        pass

    return pass
```

See comment from @cdeil in #361.

---

This is a follow-up of PR #395.

---


Code0x58:
I just thought, there is the case of decorators around classes/functions, so that regex could use and test file could use a `@` case

johanfleury:
I updated the PR and also added a commit to fix some `DeprecationWarning`.

ColinKennedy:
I just stumbled upon this PR. Is there anything that I can do to help it along or is it ready to merge?

johanfleury:
@ColinKennedy we're just waiting for a review by a project member.

Nurdok:
Sorry for the long delays in response - I've been on back-to-back family vacation and business trips. I'll to catch up as soon as I can.

src/pydocstyle/checker.py
```diff
@@ -198,7 +198,11 @@ def check_no_blank_before(self, function, docstring):  # def
             if blanks_before_count != 0:
                 yield violations.D201(blanks_before_count)
             if not all(blanks_after) and blanks_after_count != 0:
-                yield violations.D202(blanks_after_count)
+                if not (
```
Nurdok: Add a comment explaining this, possibly defining the regex in a constant or local variable with a descriptive name.

johanfleury: I completed function's docstring and added a comment on the if statement as well. I didn't find a good/useful way to name the regex so I kept it as it was.

Issue #418, opened 2019-10-17
Section title in module docstrings


raabf:
I want to have a reStructuredText header in the first line of the module docstring.

For example, a `.py` file could look like the following:
```python
"""Super Module Header
=======================

Foo bar documentation for super module.
"""
```

which would result in the following error codes:

```
/home/raabf/Dokumente/module_doc_test.py:1 at module level:
    D205: 1 blank line required between summary line and description (found 0)
/home/raabf/Dokumente/module_doc_test.py:1 at module level:
    D400: First line should end with a period (not 'r')
```
Usually the first line is a brief description of the respective element of the docstrings. That is probably a matter of taste, but I think a caption is more fitting for a module, since it is a collection of elements (functions, classes, whatever …) instead of a description of single element. If it is reStructuredText syntax, sphinx will parse that docstring and display it as a section for the elements in the file (or in an `__init_.py` for multiple files) and also add it to the TOC. That way, you can order the elements into chapters and sections.

Since titles need a second line for underlining and usually do not have a fullstop at the end, the error codes D205 and D400 appear, but they are wrong in this case. The rest of the docstrings should still apply those two rules, so I do not want to deactivate them in general.

I can imagine the following solutions (The first two would be usefuly in any way, beside this problem with section titles):

First, allow noqa in module docstrings as described in #417 . Since I have to add noqa to every module docstring, this would be inefficient, but works.

Second, beside the `--ignore` option, provide also the options `--ignore-module`,  `--ignore-class`, `--ignore-function`, …. That would allow ignoring different error codes for module docstrings, class docstrings, function docstrings, …. That would be quite nice and probably also useful in other situations.

Last, detect reStructuredText headers and deactivate the error codes D205 and D400. That is probably too specific, but would be perfect for correct checking by pydocstyle.

Jeitan:
Can I bump this?  I too write module-level docstrings this way, and it's profoundly unappealing to add noqa to every single module.  Simply turning off D205 and D400 is not an option since I still want those checks going otherwise.

KelSolaar:
I wanted to use **pydocstyle** and it is the very first thing I noticed  (and it is almost a showstopper). The `# noqa` pragma is certainly not option for us.

Freed-Wu:
Can any PR about it be accepted?

PR #427, opened 2019-10-26
Allow skip on module level messages

snudler6:
Thanks for submitting a PR!

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -18,7 +18,7 @@ New Features
 * Extend support for detecting missing argument description in Numpy style
   docstrings (#407).
 * Added support for Python 3.8 (#423).
-* Allow skipping on errors on module level docstring (#427).
+* Allow skipping errors on module level docstring via #noqa (#427).
 
 Bug Fixes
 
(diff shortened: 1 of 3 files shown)


snudler6:
resolves #418

docs/release_notes.rst
```diff
@@ -18,6 +18,7 @@ New Features
 * Extend support for detecting missing argument description in Numpy style
   docstrings (#407).
 * Added support for Python 3.8 (#423).
+* Allow skipping on errors on module level docstring (#427).
```
Nurdok: Mention the feature name (noqa), e.g.:

Allow skipping errors on module level docstring via `#noqa` (#427).

snudler6: Done.

bsamseth:
While this works for the test case, it doesn't work if there is a comment before the module docstring. An example that still reports and error:

```python
# -*- coding: utf-8 -*-
# noqa: D400
"""No period at the end""" 
```
Encoding pragmas, or `#!/path/to/python` comments are quite common, so I think this should be fixed to account for this.

PR #433, opened 2019-12-05
[DOCONLY] Document that pep257 mode disables D413

godlygeek:
Thanks for submitting a PR!


godlygeek:
D413 was removed from the pep257 convention by https://github.com/PyCQA/pydocstyle/pull/404 but the docs weren't updated.  No test is added for this since it is documentation only.  I haven't added a release notes line since it seems not worth mentioning in release notes, but I can add one if it's really needed.

PR #440, opened 2019-12-15
Skip function arguments prefixed with `_` in D417 check

### LINKED ISSUE #438: D417: Ignore arguments marked as unused
First of all thanks for a really useful tool and the great D417 which has caught many issues for me!

I feel like D417 should not complain for arguments marked as unused by prepending an underscore. This is also how pylint handles the `unused-argument` error.

Consider e.g.:
```
def handle_interrupt(self, signum: int, _frame: FrameType) -> None:
    ...
```

We provide the `_frame` argument to make the callback signature valid, but never actually use it in the later code. IMHO documenting `_frame` in this case should not be required.

samj1912:
These arguments have been marked private. As such,
it should be okay for the checker to not validate their
existence in the docstring,

Thanks for submitting a PR!

Fixes #438

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -764,6 +764,13 @@ class ConventionChecker:
             # positional argument as it is `cls` or `self`
             if definition.kind == 'method' and not definition.is_static:
                 function_args = function_args[1:]
+            # Filtering out any arguments prefixed with `_` marking them
+            # as private.
+            function_args = [
+                arg_name
+                for arg_name in function_args
+                if not is_def_arg_private(arg_name)
+            ]
             missing_args = set(function_args) - docstring_args
             if missing_args:
                 yield violations.D417(", ".join(sorted(missing_args)),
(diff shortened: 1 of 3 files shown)


samj1912:
@Nurdok review please :)

src/pydocstyle/checker.py
```diff
@@ -764,6 +764,9 @@ def _check_missing_args(docstring_args, definition):
             # positional argument as it is `cls` or `self`
             if definition.kind == 'method' and not definition.is_static:
                 function_args = function_args[1:]
+            # Filtering out any arguments prefixed with `_` marking them
+            # as private.
+            function_args = [arg for arg in function_args if not arg.startswith("_")]
```
Nurdok: Can you extract this logic to a function? e.g., `is_def_arg_private`. The code will have clearer intent (`if not is_def_arg_private(arg)`) and can be reused.

Nurdok:
Please also add a release notes line.

samj1912:
@Nurdok done. Sorry for the late update. Holiday time.

samj1912:
@Nurdok bump

Issue #234, opened 2017-02-01
How to work with long literal lines in docstring? (backslash-newline not supported)


glennmatthews:
If I have have long lines in a docstring that need to remain so (e.g., a doctest that produces long output), it gets flagged by `pycodestyle` for exceeding the 79-character line limit (`E501`). However, if I use backslash-newlines to wrap the doctest, then `pydocstyle` flags it with `D207` and `D301`. Short of using `# noqa`, is there some other workaround? If not, can `D207`/`D301` be enhanced to permit backslash-newline wrapping as valid?

example.py:

```python
"""Module docstring."""


def foo():
    """Docstring not liked by pycodestyle.

    >>> print("this is a doctest which produces some long unwrapped output, sorry about that!")
    this is a doctest which produces some long unwrapped output, sorry about that!
    """
    pass


def bar():
    """Docstring not liked by pydocstyle.

    >>> print("this is a doctest which produces some long unwrapped output,\
sorry about that!")
    this is a doctest which produces some long unwrapped output,\
sorry about that!
    """
    pass


if __name__ == "__main__":
    import doctest
    doctest.testmod()
```

```
GLMATTHE-M-33R4:cot glmatthe$ pycodestyle example.py
example.py:7:80: E501 line too long (95 > 79 characters)
example.py:8:80: E501 line too long (82 > 79 characters)
GLMATTHE-M-33R4:cot glmatthe$ pydocstyle example.py
example.py:13 in public function `bar`:
        D301: Use r""" if any backslashes in a docstring
example.py:13 in public function `bar`:
        D207: Docstring is under-indented
```

Nurdok:
I'm actually not sure what would be the best way to solve this. A "# noqa: D301" might be a reasonable solution as it looks pretty rare to me. @farmersez Any thoughts?

glennmatthews:
Is it possible to have D301 ignore backslash-newline and have D207 not check indentation of lines preceded by backslash-newline?

farmersez:
IMO, if the docstring already looks badly-formatted (not your fault, of course 😄 ), a `# noqa` should suffice. I don't think this should be a special case in the D301 check.

Nurdok:
Agreed. Closing this issue - `# noqa: D301` (or `# noqa: E501` for `pycodestyle`) is a reasonable solution.

anntzer:
[...]

PR #441, opened 2019-12-15
Fix indentation error while parsing class methods

### LINKED ISSUE #437: IndentationError on line continuations for numpydoc
as of pydocstyle 5.0.1:
```python
class T:

    def foo(self, x):
        """
        Does foo.

        Parameters
        ----------
        x : {'some', 'value', 'and', 'more', 'values', \
'numpydoc', 'says', 'they', 'have', 'to', 'fit', 'in', 'one', 'line'}
        """
```
analyzing with --convention=numpy results in
```
  File "/usr/bin/pydocstyle", line 11, in <module>
    load_entry_point('pydocstyle==5.0.1', 'console_scripts', 'pydocstyle')()
  File "/usr/lib/python3.8/site-packages/pydocstyle/cli.py", line 68, in main
    sys.exit(run_pydocstyle())
  File "/usr/lib/python3.8/site-packages/pydocstyle/cli.py", line 44, in run_pydocstyle
    errors.extend(check((filename,), select=checked_codes,
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 968, in check
    for error in ConventionChecker().check_source(source, filename,
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 123, in check_source
    for error in errors:
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 915, in check_docstring_sections
    found_numpy = yield from self._check_numpy_sections(lines, definition, docstring)
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 879, in _check_numpy_sections
    yield from self._check_numpy_section(docstring, definition, ctx)
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 693, in _check_numpy_section
    yield from cls._check_parameters_section(docstring, definition, context)
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 731, in _check_parameters_section
    yield from ConventionChecker._check_missing_args(docstring_args, definition)
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 762, in _check_missing_args
    function_args = get_function_args(definition.source)
  File "/usr/lib/python3.8/site-packages/pydocstyle/checker.py", line 1002, in get_function_args
    function_arg_node = ast.parse(textwrap.dedent(function_string)).body[0].args
  File "/usr/lib/python3.8/ast.py", line 47, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 1
    def foo(self, x):
    ^
IndentationError: unexpected indent
```

samj1912:
In order to determine the missing arguments, we currently
use ast.parse to parse partial source code. This parsing
might lead to syntax errors. We catch the syntax error and
make the parsing more resilient to errors in the source code.

Fixes #437

Thanks for submitting a PR!

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -1007,9 +1007,20 @@ def is_def_arg_private(arg_name):
     """Return a boolean indicating if the argument name is private."""
     return arg_name.startswith("_")
 
-def get_function_args(function_string):
+def get_function_args(function_source):
     """Return the function arguments given the source-code string."""
-    function_arg_node = ast.parse(textwrap.dedent(function_string)).body[0].args
+    # We are stripping the whitespace from the left of the
+    # function source.
+    # This is so that if the docstring has incorrectly
+    # indented lines, which are at a lower indent than the
+    # function source, we still dedent the source correctly
+    # and the AST parser doesn't throw an error.
+    try:
+        function_arg_node = ast.parse(function_source.lstrip()).body[0].args
+    except SyntaxError:
+        # If we still get a syntax error, we don't want the
+        # the checker to crash. Instead we just return a blank list.
+        return []
     arg_nodes = function_arg_node.args
     kwonly_arg_nodes = function_arg_node.kwonlyargs
     return [arg_node.arg for arg_node in chain(arg_nodes, kwonly_arg_nodes)]
--- a/src/tests/test_cases/sections.py
+++ b/src/tests/test_cases/sections.py
@@ -274,6 +274,25 @@ def missing_colon_google_style_section():  # noqa: D406, D407
     """
 
 
+@expect("D417: Missing argument descriptions in the docstring "
+        "(argument(s) y are missing descriptions in "
+        "'bar' docstring)", func_name="bar")
+def _test_nested_functions():
+    x = 1
+
+    def bar(y=2):  # noqa: D207, D213, D406, D407
+        """Nested function test for docstrings.
+
+        Will this work when referencing x?
+
+        Args:
+            x: Test something
+that is broken.
+
+        """
+        print(x)
+
+
 @expect(_D213)
 @expect("D417: Missing argument descriptions in the docstring "
         "(argument(s) y are missing descriptions in "
@@ -340,13 +359,15 @@ class TestGoogle:  # noqa: D203
 @expect("D417: Missing argument descriptions in the docstring "
         "(argument(s) y are missing descriptions in "
         "'test_missing_numpy_args' docstring)")
+@expect("D207: Docstring is under-indented")
 def test_missing_numpy_args(_private_arg=0, x=1, y=2):  # noqa: D406, D407
     """Toggle the gizmo.
 
     Parameters
     ----------
     x : int
-        The greatest integer.
+        The greatest integer in the history \
+of the entire world.
 
     """
 
@@ -423,5 +444,22 @@ class TestNumpy:  # noqa: D203
         ----------
         danger
             Zoneeeeee!
+
+        """
+
+
+class TestIncorrectIndent:  # noqa: D203
+    """Test class."""
+
+    @expect("D417: Missing argument descriptions in the docstring "
+            "(argument(s) y are missing descriptions in "
+            "'test_incorrect_indent' docstring)", arg_count=3)
+    def test_incorrect_indent(self, x=1, y=2):  # noqa: D207, D213, D407
+        """Reproducing issue #437.
+
+Testing this incorrectly indented docstring.
+
+        Args:
+            x: Test argument.
 
         """
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -15,6 +15,8 @@ Bug Fixes
 
 * Update convention support documentation (#386, #393)
 * Detect inner asynchronous functions for D202 (#467)
+* Fix indentation error while parsing class methods (#441).
+
 
 5.0.2 - January 8th, 2020
 ---------------------------
--- a/src/tests/test_cases/expected.py
+++ b/src/tests/test_cases/expected.py
@@ -4,13 +4,13 @@ class Expectation:
     def __init__(self):
         self.expected = set()
 
-    def expect(self, *args, arg_count=0):
+    def expect(self, *args, arg_count=0, func_name=""):
         """Decorator that expects a certain PEP 257 violation."""
         # The `arg_count` parameter helps the decorator
         # with functions that have positional arguments.
         if len(args) == 1:
             def decorate(f):
-                self.expected.add((f.__name__, args[0]))
+                self.expected.add((func_name or f.__name__, args[0]))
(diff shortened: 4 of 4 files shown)


samj1912:
@Nurdok review please :)

samj1912:
@Nurdok bump?

samj1912:
@Nurdok bump

samj1912:
@Nurdok bump

src/pydocstyle/checker.py
```diff
@@ -997,9 +997,23 @@ def get_leading_words(line):
         return result.group()
 
 
-def get_function_args(function_string):
+def get_function_args(function_definition):
     """Return the function arguments given the source-code string."""
-    function_arg_node = ast.parse(textwrap.dedent(function_string)).body[0].args
```
anntzer, line 1002: I think it may be much easier to just remove leading indentation from the first line of `function_string`, and not even bother dedenting the rest?  i.e. just replace `ast.parse(textwrap.dedent(function_string))` by `ast.parse(function_string.lstrip())` and not touch anything else.

tacaswell:
I don't think the test actually tests the issue in the original issue.

Because we need to put the full parameter specification on one line to make sphinx happy we use line continuation to wrap it, but if you indent the continuation, you end up with a whole bunch of spaces in the line.

I also do not agree that this is "incorrect" indentation for the same reason.

samj1912:
@Nurdok bump

src/pydocstyle/checker.py
```diff
@@ -997,9 +997,23 @@ def get_leading_words(line):
...
+    function_docs = function_definition.docstring
+    blank_docstring = "''''''"
+    # We are replacing the docstring with a blank docstring.
+    # This is so that if the docstring has incorrectly
+    # indented lines, which are at a lower indent than the
+    # function source, we still dedent the source correctly
+    # and the AST parser doesn't throw an error.
+    sanitized_source = function_source.replace(function_docs, blank_docstring)
+    try:
+        function_arg_node = ast.parse(textwrap.dedent(sanitized_source)).body[0].args
```
Nurdok, line 1012: I know that we already used `ast.parse` before this change, but I wonder if this works for closures? e.g.,

```.py
def foo():
    x = 1
    def bar():
        """Will this work when referencing x?"""
        print(x)
```

Can you add / make sure we have a test for this?

samj1912:
> I don't think the test actually tests the issue in the original issue.
> 
> Because we need to put the full parameter specification on one line to make sphinx happy we use line continuation to wrap it, but if you indent the continuation, you end up with a whole bunch of spaces in the line.
> 
> I also do not agree that this is "incorrect" indentation for the same reason.

The issue was caused because of "incorrectly" indented conent like https://github.com/PyCQA/pydocstyle/issues/437#issuecomment-564831245. The parameters case you mention is a subset and is covered by this.

src/pydocstyle/checker.py
samj1912: Tried this. Doesn't work

src/tests/test_cases/sections.py
```diff
@@ -274,6 +274,23 @@ def missing_colon_google_style_section():  # noqa: D406, D407
     """
 
 
+@expect("D417: Missing argument descriptions in the docstring "
+        "(argument(s) y are missing descriptions in "
+        "'bar' docstring)", func_name="bar")
+def _test_nested_functions():
+    x = 1
+
+    def bar(x=1, y=2):  # noqa: D207, D213, D406, D407
```
Nurdok, line 283: Remove `x` from `bar`'s parameters so that `bar` will be use closure.

tacaswell:
But the original issue (https://github.com/PyCQA/pydocstyle/issues/437#issue-536237028) is correct.  While I see why from a pure-parsing point of view these maybe the same, from a semantic point of view they are very different as with the `\` line continuation the second line should not be "indented".

Can you please add a test of a "correctly" formatted docstring that uses line continuation as well?

samj1912:
> But the original issue ([#437 (comment)](https://github.com/PyCQA/pydocstyle/issues/437#issue-536237028)) is correct. While I see why from a pure-parsing point of view these maybe the same, from a semantic point of view they are very different as with the `\` line continuation the second line should not be "indented".
> 
> Can you please add a test of a "correctly" formatted docstring that uses line continuation as well?

Added a test. It works as expected.

src/tests/test_cases/sections.py
samj1912: Done.

samj1912:
@Nurdok all comments should be addressed now. :)

src/pydocstyle/checker.py
samj1912: My bad - this ended up being a cleaner solution! Thanks for the idea.

samj1912:
@Nurdok bump

src/tests/test_cases/sections.py
```diff
@@ -340,13 +359,15 @@ def test_missing_args_static_method(a, x, y, _test, z=3):  # noqa: D213, D407
 @expect("D417: Missing argument descriptions in the docstring "
         "(argument(s) y are missing descriptions in "
         "'test_missing_numpy_args' docstring)")
+@expect("D207: Docstring is under-indented")
```
tacaswell, line 362: This seems like a bug that getting warning here as this is a correct docstring.

tacaswell:
Where we actually have to use line continuation in practice is in the type specification not in the body of the parameter description.  A real life example: https://github.com/matplotlib/matplotlib/blob/5445d77460f71d16dd3e1b01c4c116a741baf6b9/lib/matplotlib/axes/_axes.py#L4392-L4402

```python
class A:
    def standin(...):
        """
        ...

        Parameters
        ----------

        ...


        edgecolors : {'face', 'none', *None*} or color or sequence of color, \
default: :rc:`scatter.edgecolors`
            The edge color of the marker. Possible values:
            - 'face': The edge color will always be the same as the face color.
            - 'none': No patch boundary will be drawn.
            - A color or sequence of colors.
            For non-filled markers, the *edgecolors* kwarg is ignored and
            forced to 'face' internally.

        ...

        """
        ...

```

samj1912:
@tacaswell, I agree but this is unrelated to the original bug that this PR was made to solve. This has been a long standing issue long before #437 was reported. @anntzer has a fix for the line continuation bug at https://github.com/PyCQA/pydocstyle/pull/472

Please see our conversation there for more details. Once this PR is merged, we will merge @anntzer 's PR with this fix.

samj1912:
@Nurdok I can't merge because the release notes merge conflict dismissed your review. Can you approve again please?

src/tests/test_cases/sections.py
samj1912: This is an unrelated bug - it should be fixed by https://github.com/PyCQA/pydocstyle/pull/472 which resolves the bug https://github.com/PyCQA/pydocstyle/issues/234

PR #448, opened 2020-01-10
Fix bug with google style arg regex

### LINKED ISSUE #443: D417 - inconsistent results with Google-style (possible bug)
D417 does not report for all parameters in a function docstring. Take this example file, below.

```python

"""Something here."""

def foo(skips, verbose):
    """Do stuff.

    Args:
        skips (:attr:`.Skip`):
            Lorem ipsum dolor sit amet, consectetur adipiscing elit.
            Etiam at tellus a tellus faucibus maximus. Curabitur tellus
            mauris, semper id vehicula ac, feugiat ut tortor.
        verbose (bool):
            If True, print out as much infromation as possible.
            If False, print out concise "one-liner" information.

    """
    pass


def bar(skips, verbose):
    """Do stuff.

    Args:
        skips (:attr:`.Skip`):
            Lorem ipsum dolor sit amet, consectetur adipiscing elit.
            Etiam at tellus a tellus faucibus maximus. Curabitur tellus
            mauris, semper id vehicula ac, feugiat ut tortor.
        verbose (bool): If True, print out as much infromation as possible.
            If False, print out concise "one-liner" information.

    """
    pass

```

## Results

When I ran pydocstyle, it only complains with

```
foo.py:7 in public function `foo`:                                                                                                                                                                                                                                                            
        D417: Missing argument descriptions in the docstring (argument(s) verbose are missing descriptions in 'foo' docstring)      
```

This strikes me as inconsistent. The `verbose` parameter in `bar` is considered valid because its description starts on the same line as the parameter. `skips` doesn't start on the same line of the parameter in both `foo` and `bar` but pydocstyle doesn't ever complain about `skips`. It only seems to care about `verbose`.

Shouldn't pydocstyle be complaining that `skips` doesn't start on the same line, too?
Personally I'd prefer pydocstyle to accept that docstrings can start on the line below a parameter instead of forcing the user to write on the same like 

e.g.

This is considered valid.

```
    Args:
        verbose (bool): If True, print out as much infromation as possible.
            If False, print out concise "one-liner" information.
```

This should also be considered valid.

```
    Args:
        verbose (bool): 
            If True, print out as much infromation as possible.
            If False, print out concise "one-liner" information.
```

To me, the latter feels more "right". But to be fair though, I didn't see anything in the google style guide indicating it's okay for descriptions could start one line down. So maybe you don't agree with that suggestion. 

At the very least though, I'd like pydocstyle to enforce D417 consistently. `skips` should also be failing, which it isn't currently.

### Environment Details
Python: 3.7.3
command: `pydocstyle --ignore=D213,D202,D203,D406,D407 foo.py`
pydocstyle version: 5.0.1

samj1912:
Currently, due to the way the regex was specified, the regex
matcher was getting thrown off by types that used colons.

This change makes the regex robust to such type and fixes #443

Thanks for submitting a PR!

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -15,6 +15,8 @@ Bug Fixes
 
 * Update convention support documentation (#386, #393)
 * Detect inner asynchronous functions for D202 (#467)
+* Fix a bug in parsing Google-style argument description.
+  The bug caused some argument names to go unreported in D417 (#448).
 
 5.0.2 - January 8th, 2020
 ---------------------------
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -96,9 +96,9 @@ class ConventionChecker:
         r"(\w+)"        # Followed by 1 or more unicode chars, numbers or underscores
                         # The above is captured as the first group as this is the paramater name.
         r"\s*"          # Followed by 0 or more whitespace characters
-        r"\(?(.*?)\)?"  # Matches patterns contained within round brackets.
-                        # The `(.*?)` is the second capturing group which matches any sequence of
-                        # characters in a non-greedy way (denoted by the `*?`)
+        r"(\(.*?\))?"   # Matches patterns contained within round brackets.
+                        # The `.*?`matches any sequence of characters in a non-greedy
+                        # way (denoted by the `*?`)
         r"\s*"          # Followed by 0 or more whitespace chars
         r":"            # Followed by a colon
         ".+"            # Followed by 1 or more characters - which is the docstring for the parameter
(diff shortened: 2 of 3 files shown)


samj1912:
@Nurdok bump

samj1912:
@Nurdok bump

samj1912:
@Nurdok bump

docs/release_notes.rst
```diff
@@ -4,6 +4,13 @@ Release Notes
 **pydocstyle** version numbers follow the
 `Semantic Versioning <http://semver.org/>`_ specification.
 
+Current Development Version
+---------------------------
+
+Bug Fixes
+
+* Fix bug with google style arg regex (#448).
```
Nurdok, line 12: Capitalize Google, and make this more descriptive as to what was the original issue, e.g., "Fix a bug in parsing Google-style argument description. The bug caused some argument names to go unreported in D417 (#448)."

samj1912:
@Nurdok bump?

PR #450, opened 2020-01-18
Fix false negatives for D402

### LINKED ISSUE #284: False negatives for D402
The following function fails the D402 check:

```py
def subclasses(cls):
    """Retrieve all subclasses (direct and transitive) from `cls`.

    :param cls: The class to search.
    :type cls: ``type``

    :returns: A generator of subclasses.
    :rtype: ``iter`` of ``type``
    """
    for subclass in cls.__subclasses__():
        for subsubclass in subclasses(subclass):
            yield subsubclass
        yield subclass
```

Presumably the D402 rule just sees the method name followed by an lparen?

barakbeilin:
I want to resolve #284, I've added the appropriate test.

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -478,7 +478,10 @@ class ConventionChecker:
         """
         if docstring:
             first_line = ast.literal_eval(docstring).strip().split('\n')[0]
-            if function.name + '(' in first_line.replace(' ', ''):
+            function_signarture = function.source.split(
+                ":")[0].split('def')[-1].replace(' ', '')
+            
+            if function_signarture in first_line.replace(' ', ''):
                 return violations.D402()
 
     @check_for(Function)
@@ -701,10 +704,13 @@ class ConventionChecker:
                 except `self` or `cls` if it is a method.
 
         """
+        
         docstring_args = set()
         section_level_indent = leading_space(context.line)
         content = context.following_lines
+        
         for current_line, next_line in zip(content, content[1:]):
+            
             # All parameter definitions in the Numpy parameters
             # section must be at the same indent level as the section
             # name.
@@ -741,6 +747,7 @@ class ConventionChecker:
 
         """
         docstring_args = set()
+        
         for line in context.following_lines:
             match = ConventionChecker.GOOGLE_ARGS_REGEX.match(line)
             if match:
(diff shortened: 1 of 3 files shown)


src/tests/test_cases/test.py
```diff
@@ -316,8 +316,13 @@ def sdgfsdg23245777():
 
 
 @expect('D402: First line should not be the function\'s "signature"')
-def foobar():
-    """Signature: foobar()."""
+def foobar(foo=1, bar=1):
```
samj1912: These look good. Can you also add some tests which are false negatives to ensure we *don't* raise errors for them?

samj1912:
Can you also update the release notes with this change please?

src/tests/test_cases/test.py
barakbeilin: DONE

src/tests/test_cases/test.py
```diff
@@ -316,10 +314,20 @@ def sdgfsdg23245777():
 
 
 @expect('D402: First line should not be the function\'s "signature"')
-def foobar():
+def foobar(foo=1, bar=1):
+    """Signature: foobar(foo=1, bar=1)."""
+
```
Nurdok, line 319: This is actually an interesting case, should we also fail if the docstring contains the keyword arguments without the default value? e.g.,

```py
def foobar(foo=1, bar=1):
   """Signature: foobar(foo, bar)."""
```

samj1912: There can be lots of combinations here with args/kwargs

barakbeilin: docs.python presents functions with their default values, that's why I decided to check for this kind of signature, other might argue with my logic though. There's no explicit rule about what's a signature of a function in PEP257.

src/pydocstyle/checker.py
```diff
@@ -478,7 +478,10 @@ def check_no_signature(self, function, docstring):  # def context
         """
         if docstring:
             first_line = ast.literal_eval(docstring).strip().split('\n')[0]
-            if function.name + '(' in first_line.replace(' ', ''):
+            function_signarture = function.source.split(
```
samj1912, line 481: This might break if the function signature has type hints.

samj1912:

PR #462, opened 2020-04-15
Correctly set 'release' in sphinx config.

anntzer:
... otherwise the version in the titlebar of the rendered docs on readthedocs.org is "1.0.0".


Nurdok:
Thanks! Merging.

PR #467, opened 2020-04-28
fix: regex to catch inner functions doesn't catch asynchronous ones

matthieucan:
Hello, and thanks for the great tool!

This small PR fixes the case of detecting asynchronous inner functions, for D202 (inner classes and functions require an empty line between the outer function docstring and them). It is a follow-up of #426.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -14,6 +14,7 @@ New Features
 Bug Fixes
 
 * Update convention support documentation (#386, #393)
+* Detect inner asynchronous functions (#467)
 
 5.0.2 - January 8th, 2020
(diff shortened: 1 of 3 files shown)


docs/release_notes.rst
```diff
@@ -14,6 +14,7 @@ New Features
 Bug Fixes
 
 * Update convention support documentation (#386, #393)
+* Detect inner asynchronous functions (#467)
```
samj1912, line 17: One small comment - can you update this to say that it relates to D202?

matthieucan: Sure, done!

PR #470, opened 2020-05-04
Fix Module.is_public() when the module is not a the root

tibdex:
Thanks for submitting a PR!

### DIFF
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -1,5 +1,6 @@
 """Python code parser."""
 
+import os
 import textwrap
 import tokenize as tk
 from itertools import chain, dropwhile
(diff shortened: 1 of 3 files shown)


src/pydocstyle/parser.py
```diff
@@ -117,7 +118,8 @@ def is_public(self):
 
         This helps determine if it requires a docstring.
         """
-        return not self.name.startswith('_') or self.name.startswith('__')
+        module_name = os.path.basename(self.name)
```
Nurdok, line 121: Can you please use `pathlib` instead of `os.path`?

samj1912:
@tibdex I think this is a good patch, can you please switch to pathlib so that we can merge this :)

src/tests/parser_test.py
```diff
@@ -1,6 +1,7 @@
 """Parser tests."""
 
 import io
+import os
```
Nurdok, line 4: Please use `pathlib` in this file, too.

samj1912:
@tibdex bump?

src/tests/parser_test.py
gbroques: Updated in PR #494.

tibdex:
See https://github.com/PyCQA/pydocstyle/pull/494#issuecomment-664320247.

PR #472, opened 2020-05-07
Fix handling of dedented continuation lines.

anntzer:
- Dedent the function definition before parsing it.
- Join continued-lines before attempting to parse the docstring.

Alternative to #441; further improves handling of parameter descriptions with continued lines (which were not handled correctly before).

Thanks for submitting a PR!

- [X] Add a line to the release notes (docs/release_notes.rst) under "Current Development Version".  
   
### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -20,7 +20,7 @@ Bug Fixes
   The bug caused some argument names to go unreported in D417 (#448).
 * Fixed an issue where skipping errors on module level docstring via #noqa
   failed when there where more prior comments (#446).
-* Support backslash-continued parameter descriptions in Numpy-style docstrings
+* Support backslash-continued descriptions in docstrings
   (#472).
 
 
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -20,6 +20,8 @@ Bug Fixes
   The bug caused some argument names to go unreported in D417 (#448).
 * Fixed an issue where skipping errors on module level docstring via #noqa
   failed when there where more prior comments (#446).
+* Support backslash-continued parameter descriptions in Numpy-style docstrings
+  (#472).
 
 
 5.0.2 - January 8th, 2020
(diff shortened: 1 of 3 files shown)


samj1912:
Thanks for the patch. But from what I can see, this introduces new behaviour which is unrelated to the original issue. (It looks like if the doc string now has line continuation, it no longer errors with a D207). Although this might be a good improvement, can we please file a related issue to the repo with this behaviour and reference that instead?

Also it looks like most of the code in this PR is taken from https://github.com/PyCQA/pydocstyle/pull/441/

Can you either pick my commits from that PR to preserve the original code attribution or create a new PR with the improved handling of line continuations and update the docs to indicate this new behaviour?

samj1912:
Looks like the issue this PR is solving is -  https://github.com/PyCQA/pydocstyle/issues/234

anntzer:
The only reason I opened this separately is because you didn't pick up https://github.com/PyCQA/pydocstyle/pull/441#discussion_r414518686.  Now that you did I have no problem waiting for your PR to go in first and rebase this one over yours.

samj1912:
The original PR was fixed. Would you like to rebase and update the description with the dedented bug fix?

samj1912:
@anntzer bump?

anntzer:
fixed :)

anntzer:
What do you actually want to test for?

samj1912:
That it doesn't throw D207 for normal doc string descriptions with line continuations

anntzer:
done

docs/release_notes.rst
```diff
@@ -20,6 +20,8 @@ Bug Fixes
   The bug caused some argument names to go unreported in D417 (#448).
 * Fixed an issue where skipping errors on module level docstring via #noqa
   failed when there where more prior comments (#446).
+* Support backslash-continued parameter descriptions in Numpy-style docstrings
```
samj1912, line 23: Isn't it doing this for more than just numpy parameter descriptions at this point?

samj1912: should we update this?

anntzer:
thanks for fixing the last part :)

PR #485, opened 2020-06-09
Add disable_skip_errors flag to ConventionChecker API

### LINKED ISSUE #484: [Feature Request] Add API to disable noqa support
flake8 has an option to disable `# noqa` comments. When running pydocstyle as a flake8 plugin (via flake8-docstrings) and using flake8's `--disable-noqa` option, `# noqa` comments still block all violation reports from pydocstyle because pydocstyle is processing those comments itself and deciding to not send the reports at all.

It would be great if pydocstyle supported a feature equivalent to flake8's `--disable-noqa` and once that's in place, flake8-docstrings activated that feature so that flake8 sees all reports regardless of `# noqa` comments and could make its own decision to report them or not.

(I understand that would be a separate issue on flake8-docstrings once support is added to pydocstyle, I just mention it to ensure that the feature can be activated directly in `ConventionChecker.check_source()` as well as via traditional means).

plinss:
Added to ConventionChecker.check_source() and ConventionChecker.check().
This allows other systems that integrate pydocstyle,
such as flake8-docstyle, to request errors that are otherwise
skipped due to # noqa comments.

Added test for skip_errors feature in general and with this flag.

Fixes #484

Thanks for submitting a PR!


src/pydocstyle/checker.py
```diff
@@ -104,7 +104,7 @@ class ConventionChecker:
         ".+"            # Followed by 1 or more characters - which is the docstring for the parameter
     )
 
-    def check_source(self, source, filename, ignore_decorators=None):
+    def check_source(self, source, filename, ignore_decorators=None, disable_skip_errors=False):
```
sigmavirus24, line 107: Can we rename this `disable_skip_errors` has a certain cognitive dissonance. It almost feels like a double negative. Perhaps `ignore_inline_noqa` or `report_all_errors`?

plinss: I agree with disliking the double negative. I had originally written this as `all_errors`, but reconsidered because elsewhere in the code the feature is referred to as `skip_errors` and I thought the consistency was valuable. I had also considered `skip_errors=True`, or `allow_skip_errors=True` but in general find that non-None/False defaults tend to cause problems, and it adds a few `not`s to the logic which makes it harder to read. `skip_errors`  could also imply a list of errors to skip.

I’m happy to change it, which do you prefer?

docs/release_notes.rst
```diff
@@ -10,6 +10,7 @@ Current Development Version
 New Features
 
 * Skip function arguments prefixed with `_` in D417 check (#440).
+* Add flag to disable `# noqa` comment processing in API (#485).
```
samj1912, line 13: If we do plan on adding this to the checker - I am not sure if it makes sense to expose it only programmatically without exposing it via the CLI or the config file.

plinss: Personally I agree with you, but I didn't implement the CLI or config options because of @sigmavirus24's comments: https://github.com/PyCQA/pydocstyle/issues/484#issuecomment-640879812 https://github.com/PyCQA/pydocstyle/issues/484#issuecomment-641263791

I'm happy to add CLI and/or config support. My preference would be to add it in a separate PR as there seems to be some debate about landing that, and this change has utility regardless. See #486

Getting this merged is blocking adding support for this to flake8-docstyle which is turn is blocking https://github.com/plinss/flake8-noqa/issues/1

plinss: moved

samj1912 (approved):

samj1912:
Will release a new minor version tomorrow.

Issue #493, opened 2020-07-19
D100: Module with Single Leading Underscore Considered Public


gbroques:
**Issue:** Receive "D100: Missing docstring in public module" error for module with single leading underscore.

**pydocstyle version:** `5.0.2`

From [docs](http://www.pydocstyle.org/en/5.0.2/error_codes.html#how-publicity-is-determined):

> Publicity for all constructs is determined as follows: a construct is considered public if -

> * Its immediate parent is public and
> * **Its name does not contain a single leading underscore.**

## How to Reproduce
Consider the following simple example.
```
publicpackage/
├── __init__.py
└── _privatemodule.py
```

Within `publicpackage/__init__.py`:

```python
"""Some docstring."""
from ._privatemodule import public_function_in_private_module

__all__ = ['public_function_in_private_module']
```

Within `publicpackage/_privatemodule.py`:

```python
def public_function_in_private_module():
    """Public function in private module docstring."""
    pass
```

```
$ pydocstyle publicpackage      
publicpackage/_privatemodule.py:1 at module level:
        D100: Missing docstring in public module
```

## Expected Behavior
`_privatemodule` contains single leading underscore, and would be considered **private**, not public.

samj1912:
related pr https://github.com/PyCQA/pydocstyle/pull/470

gbroques:
> related pr #470

Opened #494 as a continuation of work in #470.

FrancescoCasalegno:
I just had the same error.

PR #494, opened 2020-07-19
Correctly detect publicity of modules inside directories

gbroques:
Continuation of PR #470 as @tibdex was taking a little to respond, and their PR relates to issue #493 that I opened.

I branched off @tibdex's work so they get contribution credits for the work they did. Thank you @tibdex!!

@samj1912 @Nurdok Please review. This uses `pathlib` in the `parser_test.py` now.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -15,7 +15,7 @@ Bug Fixes
 
 * Update convention support documentation (#386, #393)
 * Detect inner asynchronous functions for D202 (#467)
-* Correctly detect publicity of modules inside directories
+* Correctly detect publicity of modules inside directories (#470)
 
 5.0.2 - January 8th, 2020
 ---------------------------
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -17,7 +17,7 @@ Bug Fixes
 * Detect inner asynchronous functions for D202 (#467)
 * Fix a bug in parsing Google-style argument description.
   The bug caused some argument names to go unreported in D417 (#448).
-* Correctly detect publicity of modules inside directories (#470).
+* Correctly detect publicity of modules inside directories (#470, #493).
 
 5.0.2 - January 8th, 2020
 ---------------------------
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -15,6 +15,7 @@ Bug Fixes
 
 * Update convention support documentation (#386, #393)
 * Detect inner asynchronous functions for D202 (#467)
+* Correctly detect publicity of modules inside directories
 
 5.0.2 - January 8th, 2020
 ---------------------------
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -1,5 +1,6 @@
 """Python code parser."""
 
+import os
 import textwrap
 import tokenize as tk
 from itertools import chain, dropwhile
@@ -117,7 +118,8 @@ class Module(Definition):
 
         This helps determine if it requires a docstring.
         """
-        return not self.name.startswith('_') or self.name.startswith('__')
+        module_name = os.path.basename(self.name)
+        return not module_name.startswith('_') or module_name.startswith('__')
 
     def __str__(self):
         return 'at module level'
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -1,11 +1,11 @@
 """Python code parser."""
 
-import os
 import textwrap
 import tokenize as tk
 from itertools import chain, dropwhile
 from re import compile as re
 from io import StringIO
+from pathlib import Path
 
 from .utils import log
 
@@ -118,7 +118,7 @@ class Module(Definition):
 
         This helps determine if it requires a docstring.
         """
-        module_name = os.path.basename(self.name)
+        module_name = Path(self.name).name
         return not module_name.startswith('_') or module_name.startswith('__')
 
     def __str__(self):
--- a/src/tests/parser_test.py
+++ b/src/tests/parser_test.py
@@ -1,6 +1,7 @@
 """Parser tests."""
 
 import io
+import os
 import sys
 import pytest
 import textwrap
@@ -562,18 +563,22 @@ def test_matrix_multiplication_with_decorators(code):
     assert inner_function.decorators[0].name == 'a'
 
 
-def test_module_publicity():
+@pytest.mark.parametrize("parent_path", (
+    os.path.join("some", "directory"),
+    ""
+))
+def test_module_publicity(parent_path):
     """Test that a module that has a single leading underscore is private."""
     parser = Parser()
     code = CodeSnippet("")
 
-    module = parser.parse(code, "filepath")
+    module = parser.parse(code, os.path.join(parent_path, "filepath"))
     assert module.is_public
 
-    module = parser.parse(code, "_filepath")
+    module = parser.parse(code, os.path.join(parent_path, "_filepath"))
     assert not module.is_public
 
-    module = parser.parse(code, "__filepath")
+    module = parser.parse(code, os.path.join(parent_path, "__filepath"))
     assert module.is_public
 
 
--- a/src/tests/parser_test.py
+++ b/src/tests/parser_test.py
@@ -1,10 +1,10 @@
 """Parser tests."""
 
 import io
-import os
 import sys
 import pytest
 import textwrap
+from pathlib import Path
 
 from pydocstyle.parser import Parser, ParseError
 
@@ -564,21 +564,21 @@ def test_matrix_multiplication_with_decorators(code):
 
 
 @pytest.mark.parametrize("parent_path", (
-    os.path.join("some", "directory"),
-    ""
+    Path("some").joinpath("directory"),
+    Path("")
 ))
 def test_module_publicity(parent_path):
     """Test that a module that has a single leading underscore is private."""
     parser = Parser()
     code = CodeSnippet("")
 
-    module = parser.parse(code, os.path.join(parent_path, "filepath"))
+    module = parser.parse(code, str(parent_path.joinpath("filepath")))
     assert module.is_public
 
-    module = parser.parse(code, os.path.join(parent_path, "_filepath"))
+    module = parser.parse(code, str(parent_path.joinpath("_filepath")))
     assert not module.is_public
 
-    module = parser.parse(code, os.path.join(parent_path, "__filepath"))
+    module = parser.parse(code, str(parent_path.joinpath("__filepath")))
     assert module.is_public
 
 


src/tests/parser_test.py
```diff
@@ -562,18 +563,22 @@ def test_matrix_multiplication_with_decorators(code):
     assert inner_function.decorators[0].name == 'a'
 
 
-def test_module_publicity():
+@pytest.mark.parametrize("parent_path", (
+    Path("some").joinpath("directory"),
```
samj1912, line 567:
```suggestion
    Path("some") / "directory",
```

can you also replace the others please :)

gbroques: Sure thing, I'll find some time today to work on this!

gbroques: Updated in 10e1e7aad14628d52e14788aa5a5807ee9928631.

docs/release_notes.rst
```diff
@@ -20,7 +20,7 @@ Bug Fixes
   The bug caused some argument names to go unreported in D417 (#448).
 * Fixed an issue where skipping errors on module level docstring via #noqa
   failed when there where more prior comments (#446).
-
+* Correctly detect publicity of modules inside directories (#470, #493).
```
samj1912, line 23: the PR number seems incorrect?

gbroques: Woops, I had thought these were GitHub issue numbers.

Let me update this `#494`.

gbroques:
@samj1912 What's you're expectation of this test?

```python
    module = parser.parse(code, str(parent_path / "_private_pkg" / "filepath"))
    assert not module.is_public
```

My expectation is that this should pass?

Thus, a module containing no leading underscore in a private package should be considered **private**.

gbroques:
Publicity is inherently affected by where you run `pydocstyle`.

Consider the following example.

```
$ tree -F publicpackage/ ↵
publicpackage/
├── __init__.py
└── _privatepackage/
    └── some_private_module.py
``` 

```
$ pydocstyle publicpackage/ ↵
```

```
$ cd publicpackage/_privatepackage && pydocstyle some_private_module.py ↵
some_private_module.py:1 at module level:
        D100: Missing docstring in public module
```

To determine publicity correctly, we should probably be passing what the top-level package is to `pydocstyle`.

However, maybe we can simply document this limitation for now.

This was brought up in #326 in [this comment](https://github.com/PyCQA/pydocstyle/pull/326#discussion_r205946602).

@Nurdok @samj1912 thoughts?

tibdex:
I was on vacation, thanks for taking over @gbroques 👍. I'll close my PR.

gbroques:
> I was on vacation, thanks for taking over @gbroques 👍. I'll close my PR.

Not a problem! Thank you for your initial work on this.

Hopefully we can get it merged :)

src/pydocstyle/parser.py
```diff
@@ -117,7 +118,17 @@ def is_public(self):
...
+        return (
+            not self._is_inside_private_package() and
+            (not module_name.startswith('_') or module_name.startswith('__'))
+        )
+
+    def _is_inside_private_package(self):
+        """Return True if the module is inside a private package."""
+        path = Path(self.name)
+        package_names = path.parts[:-1]
+        return any([package.startswith('_') for package in package_names])
```
samj1912, line 131: How does this handle any directories starting with `__`?

gbroques: > How does this handle any directories starting with `__`?

It considers them **private**, and I added a unit test for this in 84fa77f07ecd2d7373a793d1130ec61a75a14bac.

Should they be considered **public** instead like modules starting with double-underscores?

Nurdok: Modules with double underscore should be considered **public** (e.g., `__main__.py`, `__init__.py`, etc.)

gbroques: Thanks for the clarification @Nurdok.

Considering *special* "dunder" modules like `__init__.py` **public** makes sense.

Do you consider modules like `__my_module.py`, *without* the *trailing* double-underscores,  **public** or **private**?

samj1912: I would keep any module that begins and ends with a double underscore public and the rest should be considered private. So something like `__main__.py` should be public but something like `__my_module.py` would hopefully be private.

src/pydocstyle/parser.py
```diff
@@ -117,7 +118,25 @@ def is_public(self):
...
+
+    def _is_inside_private_package(self):
+        """Return True if the module is inside a private package."""
+        path = Path(self.name)
+        package_names = path.parts[:-1]
+        return any([self._is_private_name(package) for package in package_names])
+
+    def _is_public_name(self, module_name):
+        """Determine whether a "module name" (i.e. module or package name) is public."""
+        return not module_name.startswith('_') or module_name.startswith('__')
```
samj1912, line 135: I think we should modify it to start *and* end with double underscores for it to be private. thoughts?

gbroques: Sorry, can you elaborate on what you mean by this?

>  So something like `__main__.py` should be **public** but something like `__my_module.py` would hopefully be **private**.

If this is what you mean, then I'm re-working the code now. :+1: 

If you meant something else, like re-factoring the code to express the logic in a different way, then please provide code-snippets or more explanation.

src/pydocstyle/parser.py
gbroques: > So something like `__main__.py` should be **public** but something like `__my_module.py` would hopefully be **private**.

Updated in https://github.com/PyCQA/pydocstyle/pull/494/commits/ef7f5a1945e2a20f42e5913796892abc478656f9.

src/tests/parser_test.py
```diff
@@ -562,20 +563,70 @@ def test_matrix_multiplication_with_decorators(code):
...
+    # -----------------
+    module = parser.parse(code, str(parent_path / "__private_module"))
+    assert not module.is_public
+
+    module = parser.parse(code, str(
+        parent_path / "__private_package" / "module"))
+    assert not module.is_public
+
+    module = parser.parse(code, str(
+        parent_path / "__private_package" / "package" / "module"))
```
samj1912, line 627: Shouldn't all these file paths end with `.py`? (I know that the code will correctly handle it since we call Path.stem , but just saying since someone might modify it in the future and since we do check for endswith :) )

src/tests/parser_test.py
```diff
@@ -562,20 +563,70 @@ def test_matrix_multiplication_with_decorators(code):
     assert inner_function.decorators[0].name == 'a'
 
 
-def test_module_publicity():
-    """Test that a module that has a single leading underscore is private."""
+@pytest.mark.parametrize("parent_path", (
+    Path("package") / "another_package",
+    Path("")
+))
+def test_module_publicity(parent_path):
```
samj1912, line 570: I think it's also worth parametrizing all the examples below. That will significantly improve the readability of this test code and will make sure all of them are run independently.

You can modify the mark to hold file path and the expected publicity and simply check that below instead of manually repeating all the parse calls and asserts :)

src/tests/parser_test.py
gbroques: Good point @samj1912.

I can update this. :)

src/tests/parser_test.py
gbroques: Good suggestion! I will re-work it, and notify when finished. :)

gbroques: @samj1912 Let me know if this addresses your comments https://github.com/PyCQA/pydocstyle/pull/494/commits/cec57934bf781f3d4956bdcc2c8c62bd6123cc73.

> You can modify the `mark.parametrize` to hold the file path **and the expected publicity**.

If we want a single `test_module_publicity` instead of two separate ones for **public** and **private**, then I'm happy to refactor that as well.

src/tests/parser_test.py
gbroques: Added `.py` file extensions to test path examples in https://github.com/PyCQA/pydocstyle/pull/494/commits/cec57934bf781f3d4956bdcc2c8c62bd6123cc73.

samj1912:
lgtm, thanks for the patch. I will wait for @Nurdok to get a chance to look as well. :)

gbroques:
@samj1912 Sure, thanks for helping maintain and support this library! 🙂

src/pydocstyle/parser.py
```diff
@@ -117,7 +118,29 @@ def is_public(self):
...
+        module_name = Path(self.name).stem
+        return (
+            not self._is_inside_private_package() and
+            self._is_public_name(module_name)
+        )
+
+    def _is_inside_private_package(self):
+        """Return True if the module is inside a private package."""
+        path = Path(self.name)
+        package_names = path.parts[:-1]
```
Nurdok, line 130: I think there's a fundamental issue here. The publicity of a module doesn't depend on its _entire_ path, just the path leading to directory inside one of the paths in `PYTHONPATH`.

So let's say I have a module in path `/_foo/bar/baz.py`. If `PYTHONPATH` is set to `/`, then `baz.py` is private. If `PYTHONPATH` is set to `/_foo/` then `baz.py` is public.

Fortunately `sys.path` is a list of all the filepaths in `PYTHONPATH`, so this can be relatively simply implemented. Something like:

```py
def _is_inside_private_package(name):
    path = Path(name).parent  # Ignore the actual module's name
    syspath = [Path(p) for p in sys.path]  # Convert to pathlib.Path.

    while path != path.parent and path not in syspath:  # Bail if we are at the root directory or in `PYTHONPATH`.
        if _is_private_name(path.name):
            print(f"Stopping at private package: {path}")
            return True
        path = path.parent

    return False
```

gbroques: Good catch! Let me try and incorporate your suggestion, and ideally add a unit test to capture this.

gbroques: Please let me know if https://github.com/PyCQA/pydocstyle/pull/494/commits/4b57b42da6007dba708ad0052189a9e0bebda426 addresses your comment, and if there are any other changes desired.

src/pydocstyle/parser.py
```diff
@@ -117,7 +119,35 @@ def is_public(self):
...
+            not self._is_inside_private_package() and
+            self._is_public_name(module_name)
+        )
+
+    def _is_inside_private_package(self):
+        """Return True if the module is inside a private package."""
+        path = Path(self.name).parent  # Ignore the actual module's name
+        syspath = [Path(p) for p in sys.path]  # Convert to pathlib.Path.
+
+        while path != path.parent and path not in syspath:  # Bail if we are at the root directory or in `PYTHONPATH`.
```
Nurdok, line 133: Sorry for the error in my own code, but I guess this exceeds 80 chars? Just move the comment one line up.

src/tests/parser_test.py
```diff
@@ -562,19 +563,57 @@ def test_matrix_multiplication_with_decorators(code):
...
+    Path("_private_package") / "package" / "module.py",
+    Path("") / "_private_package" / "package" / "module.py",
+
+    # double underscore
+    Path("__private_module.py"),
+    Path("__private_package") / "module.py",
+    Path("__private_package") / "package" / "module.py",
+    Path("") / "__private_package" / "package" / "module.py"
+))
+def test_module_publicity_with_private_paths(private_path):
```
Nurdok, line 603: Please also add some tests with the same path over different `sys.path` cases.

gbroques: Can you elaborate on this please?

Should I add new tests that modify or mutate `sys.path` in the test, or just add more test path examples to the existing two tests?

gbroques: Can you elaborate on this please?

Should I add new tests that modify or mutate `sys.path` in the test, or just add more test path examples to the existing two tests?

src/pydocstyle/parser.py
gbroques: No worries, this should be resolved in https://github.com/PyCQA/pydocstyle/pull/494/commits/f8073657a6b0a63c89d33b2c2a7b2a9ffe25e306.

docs/snippets/publicity.rst
```diff
@@ -10,7 +10,9 @@ Publicity for all constructs is determined as follows: a construct is
 considered *public* if -
 
 1. Its immediate parent is public *and*
```
samj1912, line 12: Should we also update the docs to include our new behavior?

src/pydocstyle/parser.py
```diff
@@ -117,7 +119,36 @@ def is_public(self):
...
+            self._is_public_name(module_name)
+        )
+
+    def _is_inside_private_package(self):
+        """Return True if the module is inside a private package."""
+        path = Path(self.name).parent  # Ignore the actual module's name
+        syspath = [Path(p) for p in sys.path]  # Convert to pathlib.Path.
+
+        # Bail if we are at the root directory or in `PYTHONPATH`.
+        while path != path.parent and path not in syspath:
```
samj1912, line 134: I don't think we should iterate all the way to the root. But only till the current directory? Thoughts?

samj1912: edit - nvm, the current working directory should always be in the sys path so we should be fine.

docs/snippets/publicity.rst
gbroques: Which new behavior are you referring to?

samj1912: Ah - I meant the behavior around syspath. That may catch some people off guard.

src/pydocstyle/parser.py
gbroques: I did some tests with the `pydocstyle` CLI locally, and the directory that you execute `pydocstyle` from is **not included** in `sys.path`.

> I don't think we should iterate all the way to the root. But only till the current directory? 

Maybe your right? Please confirm and advise.

docs/snippets/publicity.rst
gbroques: It's kind of an internal implementation detail, but I can make mention of it as a "Note" or something near the bottom of the "How publicity is determined" section.

> **Note**, a construct's parent is recursively checked upward until we reach a directory in `sys.path` to avoid considering the complete filepath.

Thoughts?

src/pydocstyle/parser.py
samj1912: I am just wondering about this use case because if we iterate all they way till root, there might be folders with an underscore in their name. On the other hand this means that pydocstyle will have different outputs on the same input based on the `CWD` which I am not sure about. So - let's keep it to `sys.path` for now simply because it makes the checker consistent.

In the future if we decide to change this behavior we can always put in a PR - @Nurdok what do you think?

docs/snippets/publicity.rst
samj1912: Let's keep this comment to what you suggested for now.

gbroques:
> @gbroques Thank you for being so persistent and patient about this PR. I think this is good to go but I would like @Nurdok to take a look before merging since he had the last significant review on this PR.

No worries! I really appreciate all the excellent feedback you and @Nurdok have provided, and this is just a small way I can say thank you for the work you two do. :)

docs/snippets/publicity.rst
gbroques: Updated in commit https://github.com/PyCQA/pydocstyle/pull/494/commits/b72179dff7feec5e0880414d7e56a42b15b3a952.

PR #499, opened 2020-08-23
Adding tests for parsing one line functions/ Fixing parser bug.

theyuvalraz:


src/tests/parser_test.py
```diff
@@ -888,6 +888,12 @@ def test():
         def test():
             pass
     """),
+    CodeSnippet("""\
+        '''Test this'''
+
+        @property
```
samj1912, line 894: Can you also add some tests for multiple decorators applied to single line function?

samj1912 (changes requested):

src/tests/parser_test.py
theyuvalraz: Sure, Added this scenario in the parameterised test.
without adding a new test i can't test that both decorators are recorded, will add it later if you think it's needed.

src/tests/test_cases/test.py
```diff
@@ -337,6 +337,16 @@ def oneliner_d102(): return
 def oneliner_withdoc(): """One liner"""
 
 
+def ignored_decorator(func):   # noqa: D400,D401,D415
+    """Runs something"""
+    func()
+    pass
+
+
+@ignored_decorator
```
samj1912, line 346: Can you please add a test case next to this one with a single line definition similar to this one that doesn't have an ignored decorator and raises errors and another one that has a single line definition with a docstring that doesn't raise any errors to check that everything is working as expected in this context?

theyuvalraz: Good idea :)
Are the tests I added meet the criteria?

samj1912 (approved):
lgtm, thanks for the PR!

samj1912:
@theyuvalraz I see that we forgot to update the release notes for this PR. Can you please add a separate PR that updates them and references this one?

PR #507, opened 2020-08-29
Blacken the source code and sort imports

samj1912:
Thanks for submitting a PR!

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -5,20 +5,36 @@ import string
 import sys
 import textwrap
 import tokenize as tk
-from itertools import takewhile, chain
-from re import compile as re
 from collections import namedtuple
+from itertools import chain, takewhile
+from re import compile as re
 
 from . import violations
 from .config import IllegalConfiguration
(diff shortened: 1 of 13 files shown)


src/pydocstyle/checker.py
```diff
@@ -91,42 +108,55 @@ class ConventionChecker:
     # " random         : test" where random will be captured as the param
     # "  random_t (Test) : test  " where random_t will be captured as the param
     GOOGLE_ARGS_REGEX = re(
-                        # Matches anything that fulfills all the following conditions:
```
Nurdok, line 94: Can you move each comment to the line above the regex it describes? This hurts readability IMO.

Nurdok (changes requested):

src/pydocstyle/checker.py
samj1912: Fixed :)

samj1912:
@Nurdok bump

Nurdok (approved):

Nurdok:
Sorry for the delay @samj1912. Between the first review and now, my wife managed to give birth to a healthy baby boy :)

PR #509, opened 2020-08-30
Test on Python 3.9-dev

hugovk:
Python 3.9 is out in October, with rc1 now available, and it's strongly recommended to test against it now to make sure libraries work, and also to help test 3.9 itself.

There was an error running mypy:

```
py39 run-test: commands[1] | mypy --config-file=tox.ini src/
src/pydocstyle/utils.py:25: error: syntax error in type comment
Found 1 error in 1 file (checked 35 source files)
ERROR: InvocationError for command /private/tmp/pydocstyle/.tox/py39/bin/mypy --config-file=tox.ini src/ (exited with code 2)
```

This looks like it was fixed in mypy in https://github.com/python/mypy/pull/8716, and released in mypy v0.780. So let's upgrade to use the newest 0.782.

---

Thanks for submitting a PR!

- [n/a?] Add a line to the release notes (docs/release_notes.rst) under "Current Development Version".  
   
### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,12 +8,12 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: [3.5, 3.6, 3.7, 3.8]
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9-dev]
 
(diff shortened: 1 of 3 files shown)


samj1912 (approved):
Thanks for the PR!

PR #510, opened 2020-08-31
Drop support for Python 3.5

### LINKED ISSUE #508: Drop support for python 3.5
Python 3.5 is EOL on 13 September 2020. We should drop support for it in the next release

hugovk:
Fixes #508.

Also upgrades Python syntax with `pyupgrade --py36-plus`

And removes D302, which is redundant and cannot be run now Python 2 is dropped.

---

Thanks for submitting a PR!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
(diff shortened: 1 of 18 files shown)


samj1912:
I think the first one makes more sense as it also leaves some clues as to why it was deprecated.

hugovk:
Thanks, updated!

samj1912 (approved):

PR #511, opened 2020-09-05
Preventing the D103 error when the function is decorated with @overload.

### LINKED ISSUE #419: Wrong “D103: Missing docstring in public function” for overload functions
For example:

```python
from typing import overload

@overload
def func(a: int) -> str:
     ...
@overload
def func(a: str) -> str:
     ...
def func(a):
     """Foo bar documentation."""
     return str(a)
```

will result in 

```
/home/raabf/Dokumente/overload_test.py:4 in public function `func`:
/home/raabf/Dokumente/overload_test.py:7 in public function `func`:
```

Which is wrong since only the last def should have a docstring.

Two things should be done:

First, prevent the error D103 for overload functions. We could make use of the `--ignore-decorators='overload'`, but since it is universally valid, it should be a fixed part of it, and the user should not set it by himself.

Second, since `@overload` functions should never have docstrings, it would be useful if there is an additional error code in pydocstyle, if someone is setting a docstring for an `@overload` function.

theyuvalraz:
Fixes #419 

Added an is_overload method in the function class(parser.py).

Added an if statement so that the D103 error will not trigger when decorated with @overload(checker.py)

Added some tests to see that it's working correctly.

The lambda method in the checker returns None, not sure its a good practice. any suggestions ?


samj1912:
@theyuvalraz can you please rebase :)?

samj1912:
@theyuvalraz we also introduced isort and black formatting to the source code. 

You will want to run black src/pydocstyle and isort src/pydocstyle

after installing the pinned versions from the requirements file.

src/tests/test_cases/test.py
```diff
@@ -61,6 +61,8 @@ def overloaded_func(a: int) -> str:
 
 @overload
 def overloaded_func(a: str) -> str:
+    # TODO(Find a way to test D418
+    #  as @overload can't be passed to the @expect function)
     return '1'
```
samj1912, line 66: overloaded functions shouldn't have a definition

theyuvalraz: Right, didn't know that.

src/tests/test_cases/test.py
```diff
@@ -53,6 +54,23 @@ def nested():
...
 
 
+@overload
+def overloaded_func(a: int) -> str:
+    ...
+
+
+@overload
+def overloaded_func(a: str) -> str:
+    # TODO(Find a way to test D418
```
samj1912, line 64: Can we test this somehow?

Maybe add an integration test and see if it raises the right error?

samj1912: See https://github.com/PyCQA/pydocstyle/blob/master/src/tests/test_integration.py

theyuvalraz: Looks like we can :)

src/tests/test_cases/test.py
```diff
@@ -53,6 +54,21 @@ def nested():
         ''
 
 
+@overload
+def overloaded_func(a: int) -> str:
+    ...
+
+
+@overload
+def overloaded_func(a: str) -> str:
```
samj1912, line 63: What happens here if you do add a docstring?

theyuvalraz: You would get the expected D418 error.

theyuvalraz: How about that, I was sure that the @expect decorator is the only way to expect in this module, I was wrong...

src/pydocstyle/checker.py
```diff
@@ -213,8 +213,12 @@ def check_docstring_missing(self, definition, docstring):
                         else violations.D102()
                     )
                 ),
-                Function: violations.D103,
                 NestedFunction: violations.D103,
+                Function: (
```
samj1912, line 217: We shoudl also add support for method overloads and nested function overloads here?

theyuvalraz: Hmmm, definitely support for methods is needed. Probably nested too.

theyuvalraz: Well, this will be a bit of work to add those, I will have fun doing that over the weekend :)

src/tests/test_integration.py
```diff
@@ -502,6 +501,66 @@ def foo():
...
+        '''))
+    env.write_config(ignore="D100")
+    out, err, code = env.invoke()
+    assert code == 1
+    assert 'D418' in out
+    assert 'D103' not in out
+
+
+def test_overload_nested_function(env):
+    """Nested functions decorated with @overload trigger D418 error."""
```
samj1912, line 535: Can we also add a test here for overloaded methods?

src/pydocstyle/checker.py
```diff
@@ -204,17 +202,23 @@ def check_docstring_missing(self, definition, docstring):
...
-                        violations.D107()
-                        if definition.is_init
-                        else violations.D102()
+                        violations.D102()
+                        if not definition.is_overload
+                        else None
                     )
                 ),
-                Function: violations.D103,
                 NestedFunction: violations.D103,
```
samj1912, line 216: I'm surprised no changes were needed here?

src/tests/test_integration.py
```diff
@@ -502,6 +501,66 @@ def foo():
             in err)
 
 
+def test_overload_function(env):
+    """Functions decorated with @overload trigger D418 error."""
+    with env.open('example.py', 'wt') as example:
```
samj1912, line 506: Can we also add valid test cases where D103 is not raised?

samj1912 (changes requested):

src/pydocstyle/checker.py
theyuvalraz: Yes, me too but it looks like a "nested function" is a function that isn't public.
as its composed from the original function class then no changes are needed.
Well, I think this is the case, not an expert on python, but it works :)

src/tests/test_integration.py
```diff
@@ -502,6 +501,159 @@ def foo():
...
+                """Foo bar documentation."""
+                return str(a)
+
+        '''))
+    env.write_config(ignore="D100, D203")
+    out, err, code = env.invoke()
+    assert code == 0
+
+
+def test_overload_function_valid(env):
```
samj1912, line 597: Also a similar one for nested functions?

samj1912:
apart from the above lgtm

src/tests/test_integration.py
theyuvalraz: Sure :)

samj1912:
Thanks everything looks good! Can you also update the release notes with this change?

samj1912 (approved):

docs/release_notes.rst
```diff
@@ -20,6 +20,16 @@ Bug Fixes
 
 * Treat "package" as an imperative verb for D401 (#356).
 
+5.1.2 - September 13th, 2020
```
samj1912, line 23: Can you put this under the current development version please? I would like to update the version number and release date when I will actually release it :)

docs/release_notes.rst
```diff
@@ -20,6 +20,16 @@ Bug Fixes
...
+----------------------------
+
+New Features
+
+* Methods, Functions and Nested functions that have a docstring now throw D418 (#511).
+* Methods decorated with @overload no longer reported as D102.
```
samj1912, line 29: This and the one below should be tagged with your PR number as well :)

docs/release_notes.rst
theyuvalraz: I will try, it's my first time updating release notes.

samj1912 (approved):

samj1912:
Thank you for this pull request!

domdfcoding:
@samj1912 would it be possible to have a new release of `pydocstyle` with this feature?

samj1912:
Yeah, didn't notice this was not released. I will cut a release later today.

PR #534, opened 2021-05-01
Enable full `toml` configuration and `pyproject.toml`

### LINKED ISSUE #447: pyproject.toml configuration support
Many other python formatting/style tools have added support for the `pyproject.toml` file's `[tool.*]` sections for their configuration. Is this something that pydocstyle would be open to supporting?

RuRo:
This PR closes #447.

### Changes

- Fix a small unrelated bug, where inline comments in `.ini` files behaved differently when autodetected by `_get_config_file_in_folder` and when manually specified by `--config`.
- Add `toml` as a package dependency (`setup.py`) and as a pinned runtime dependency (`requirements/runtime.txt`).
- Implement `TomlParser` - a thin wrapper around the `toml` library with minimal API parity with `RawConfigParser`.
- Autodetect `pyproject.toml` configuration
- Autodetect `tool.pydocstyle` section
- Implement error code list support for `toml`. This is so that you can use

  ```toml
  select = [
      "D100",
      # This line is ignored
      "D103",
  ]
  ```
  and not only

  ```toml
  select = "D100,D103"
  ```
- Almost all tests in `test_integration.py` which used the `env` fixture now are run for both `toml` and `ini` versions of the configuration (47 tests).
- 2 tests are only run for `ini` configuration: `test_ignores_whitespace_in_fixed_option_set` and `test_multiple_lined_config_file`. In `toml`, strings must be quoted, so comments and newlines **inside** the strings don't make much sense, which is what these tests are checking. Instead, I've implemented 2 "equivalent" `toml` tests that also check the list support: `test_accepts_ignore_error_code_list` and `test_accepts_select_error_code_list`.

### Please make sure to check for the following items:

There's currently no "Current Development Version" section in `docs/release_notes.rst`. Where should I add the changes?

### DIFF
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -2,10 +2,12 @@
 
 import copy
 import itertools
+import operator
 import os
 from collections import namedtuple
 from collections.abc import Set
 from configparser import NoOptionError, NoSectionError, RawConfigParser
+from functools import reduce
 from re import compile as re
 
 import toml
@@ -29,7 +31,7 @@ class TomlParser:
     """ConfigParser that partially mimics RawConfigParser but for toml files.
 
     See RawConfigParser for more info. Also, please note that not all
-    RawConfigParser functionality is implemented, but only the subset, that is
+    RawConfigParser functionality is implemented, but only the subset that is
     currently used by pydocstyle.
     """
 
@@ -64,13 +66,14 @@ class TomlParser:
         return read_ok
 
     def _get_section(self, section, allow_none=False):
-        current = self._config
-        for p in section.split('.'):
-            if isinstance(current, dict) and p in current:
-                current = current[p]
-            else:
-                current = None
-                break
+        try:
+            current = reduce(
+                operator.getitem,
+                section.split('.'),
+                self._config,
+            )
+        except KeyError:
+            current = None
 
         if isinstance(current, dict):
             return current
@@ -86,8 +89,6 @@ class TomlParser:
     def options(self, section):
         """Return a list of option names for the given section name."""
         current = self._get_section(section)
-        # current = current.copy()
-        # current.update(self._defaults)
         return list(current.keys())
 
     def get(self, section, option, *, _conv=None):
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -100,6 +100,11 @@ class TomlParser:
         except KeyError:
             raise NoOptionError(option, section)
 
+        if isinstance(value, dict):
+            raise TypeError(
+                f"Expected {section}.{option} to be an option, not a section."
+            )
+
         # toml should convert types automatically
         # don't manually convert, just check, that the type is correct
         if _conv is not None and not isinstance(value, _conv):
@@ -654,7 +659,7 @@ class ConfigurationParser:
             file.
 
             """
-            if not isinstance(value_str, list):
+            if isinstance(value_str, str):
                 value_str = value_str.split(",")
             return cls._expand_error_codes(
                 {x.strip() for x in value_str} - {""}
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -184,7 +184,7 @@ class ConfigurationParser:
         '.pep257',
     )
 
-    POSSIBLE_SECTION_NAMES = ('pydocstyle', 'pep257')
+    POSSIBLE_SECTION_NAMES = ('pydocstyle', 'pep257', 'tool.pydocstyle')
 
     def __init__(self):
         """Create a configuration parser."""
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -5,8 +5,10 @@ import itertools
 import os
 from collections import namedtuple
 from collections.abc import Set
-from configparser import RawConfigParser
+from configparser import NoOptionError, NoSectionError, RawConfigParser
 from re import compile as re
+
+import toml
 
 from .utils import __version__, log
 from .violations import ErrorRegistry, conventions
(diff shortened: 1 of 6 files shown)


src/tests/test_integration.py
```diff
@@ -48,10 +55,23 @@ def write_config(self, prefix='', name='tox.ini', **kwargs):
         if not os.path.isdir(base):
             self.makedirs(base)
 
+        name = self.config_name if name is None else name
+        if name.endswith('.toml'):
+            def convert_value(val):
+                if isinstance(val, bool):
+                    return {True: 'true', False: 'false'}[val]
+                else:
+                    return repr(val)
```
Nytelife26, line 64:
```suggestion
                return repr(val).lower()
```
This can be made simpler too

src/pydocstyle/config.py
```diff
@@ -23,6 +25,98 @@ def _decorator(self, *args, **kwargs):
     return _decorator
 
 
+class TomlParser:
+    """ConfigParser that partially mimics RawConfigParser but for toml files.
+
+    See RawConfigParser for more info. Also, please note that not all
+    RawConfigParser functionality is implemented, but only the subset, that is
```
Nytelife26, line 32:
```suggestion
    RawConfigParser functionality is implemented, but only the subset that is
```
Small grammar nitpick coming through

src/pydocstyle/config.py
```diff
@@ -23,6 +25,98 @@ def _decorator(self, *args, **kwargs):
...
+        return read_ok
+
+    def _get_section(self, section, allow_none=False):
+        current = self._config
+        for p in section.split('.'):
+            if isinstance(current, dict) and p in current:
+                current = current[p]
+            else:
+                current = None
+                break
```
Nytelife26, line 73:
```suggestion
        try:
            current = reduce(operator.getitem, section.split('.'), self._config)
        except KeyError:
            current = None
```
Using `operator.getitem` and `functools.reduce` just seems cleaner to me. Less verbose, easier to read. Personal preference on that one though.

Nytelife26 (changes requested):

src/tests/test_integration.py
RuRo: This will also convert `"D100"` to `"d100"` which is wrong. Although I **can** change the `dict` lookup in the first branch to `repr(val).lower()` in the first branch only.

src/pydocstyle/config.py
```diff
@@ -552,8 +653,10 @@ def _get_set(value_str):
             file.
 
             """
+            if not isinstance(value_str, list):
```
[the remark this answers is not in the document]
RuRo: Well my intention here was so that if `value_str` is for some reason an iterable that is not a `list`, then it should fail instead of silently iterating over it. A concrete example would be something like:
```toml
[tool.pydocstyle.add_ignore]
D100 = true
D200 = false
```
This is an invalid configuration, which would (I think) fail in my version when trying to split the `{'D100': True, 'D200': False}` dictionary value and will silently iterate over the keys in your version.

Although this is a stupid situation in the first place, so I'm not too opposed to changing it to your version.

src/tests/test_integration.py
Nytelife26: Interesting, I forgot about that. Perhaps

```py
return repr(val) if not isinstance(val, bool) else repr(val).lower()
```
would suffice?

src/pydocstyle/config.py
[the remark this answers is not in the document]
Nytelife26: If that's what you're trying to do, I recommend being more explicit in throwing a proper error rather than just leaving nature of fault to deal with it.

```py
if isinstance(value_str, str):
    value_str = value_str.split(",")
elif isinstance(value_str, list):
    pass
else:
    throw ErrorGoesHere("message goes here")
```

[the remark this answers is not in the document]
RuRo: I moved the check for the case I was worried about to `TomlParser.get` instead.

RuRo:
@Nurdok, sorry to bother you, but can you maybe take a look at this PR?

Nurdok:
@samj1912 is the current active maintainer - I'll let him address this PR.

samj1912:
Will take a look this weekend :)

samj1912:
This generally looks good to me, just a couple of comments -

1. Can we make toml an optional dependency so that if it's not present, people can still pydocstyle. This is so that you can install pydocstyle with toml support using something like pydocstyle[toml]. If toml is not present and pyproject.toml is, pydocstyle could show a warning that it maybe missing configuration and that users should install toml to get support for it. 
2. Can we have the section names conditional to the file being parsed? I would prefer if we only parsed the tool.pydocstyle section in pyproject.toml and not the other top level sections.

samj1912:
> 1. Theoretically we can, however it's not clear to me, what should happen if the `pyproject.toml` file is present, but `toml` is not installed.
>    
>    * Raising an error might break existing projects that currently have a `pyproject.toml` file and either don't have a `.pydocstyle` or have it in a subdirectory instead of in the root of the project.
>    * Ignoring the `pyproject.toml` file if `toml` is not installed can lead to unexpected results, since `toml` can also be installed as a dependency of some other package or can be installed on the developers machine, but not in CI or vice-versa.

I meant something similar to this - https://github.com/cruft/cruft/pull/107/files

> 2. Done. Just to be clear,
>    > I would prefer if we only parsed the tool.pydocstyle section in pyproject.toml and not the other top level sections.
>    
>    
>    `toml` will still technically **parse** the whole `pyproject.toml` file no matter what. I just changed it so that the `pydocstyle` configuration keys are only looked up in `tool.*` instead of also checking for `pydocstyle`/`pep257` in the root.

Yup, this is what I meant :).

RuRo:
> I meant something similar to this - https://github.com/cruft/cruft/pull/107/files

Sure, I guess a warning is an acceptable compromise.

samj1912:
Code wise it all looks good :) Could you also update the changelog with an entry for this PR and if possible update the docs! Thanks

RuRo:
> Could you also update the changelog with an entry for this PR

As I mentioned in the original post

> There's currently no "Current Development Version" section in docs/release_notes.rst. Where should I add the changes?

samj1912:
Thank you for this PR :)

RuRo:
👍

Btw, I was wondering, does `pydocstyle` have a release schedule? Do you have any estimations of when might the next version be released? I'm itching to bump the version in my personal projects :^)

samj1912:
It does not have a fixed release schedule. I can cut a release sometime next week though :)

Nytelife26:
Nice work :) turns out it wasn't at all as complex as implied, eh. That, or you're a software engineer from beyond modern earth.

All jokes aside, thank you for this - I'll be using this feature in proselint shortly too

PR #536, opened 2021-05-09
fix `--source`

### LINKED ISSUE #535: `--source` is broken
The `--source` option doesn't work correctly. For example:
```python
def foo():
    pass
```

```
> pydocstyle test.py --source

test.py:1 at module level:
        D100: Missing docstring in public module

 1: d 2: e 3: f 4:   5: f 6: o 7: o        ...


test.py:1 in public function `foo`:

 1: d 2: e 3: f 4:   5: f 6: o 7: o        ...
```

The problem is that [`Error.lines` in `pydocstyle/violations.py`](https://github.com/PyCQA/pydocstyle/blob/837c0c2ce7481fb20ae9af115203348961734e8c/src/pydocstyle/violations.py#L61) expects [`Definition.source` defined in `pydocstyle/parser.py`](https://github.com/PyCQA/pydocstyle/blob/837c0c2ce7481fb20ae9af115203348961734e8c/src/pydocstyle/parser.py#L118) to be a list of strings, but it's actually a single string with newlines in it. So instead of iterating over lines, it ends up iterating over characters in that string.

RuRo:
This PR closes #535.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -8,6 +8,9 @@ Release Notes
 Current Development Version
 ---------------------------
 
+Bug Fixes
+
+* Split ``--source`` by lines instead of by characters (#536).
 
 6.1.0 - May 17th, 2021
 ---------------------------
--- a/src/pydocstyle/violations.py
+++ b/src/pydocstyle/violations.py
@@ -63,7 +63,7 @@ class Error:
         if self.definition is None:
             return ''
         source = ''
-        lines = self.definition.source
+        lines = self.definition.source.splitlines(keepends=True)
         offset = self.definition.start  # type: ignore
         lines_stripped = list(
             reversed(list(dropwhile(is_blank, reversed(lines))))


RuRo:
@samj1912 any comments for this PR? Also, you mentioned that you could make a new release soon, any progress on that?

docs/release_notes.rst
```diff
@@ -20,6 +20,10 @@ New Features
 
 * Enable full toml configuration and pyproject.toml (#534).
 
+Bug Fixes
```
samj1912, line 23: Can you move this up to Current dev version?

RuRo: Done

samj1912 (approved):

RuRo:
Sorry to bother you, but would you mind also releasing this fix right away while we're at it? I kind of originally expected to get this merged into `6.1.0`.

PR #541, opened 2021-05-17
Fix CI for python 3.6

### LINKED ISSUE #540: Investigate fixing CI for python 3.6
> Python 3.6 has not been getting bug fixes since 3.6.8 (which was released in 2018). The latest tests were failing on 3.6 on macos. I do not want to block PRs on it and I do not have the time to put in a proper fix for such an old version of Python (released 5 years ago). I have currently staged this removal for the next major release of pydocstyle. If someone wants to fix the tests and add back support for 3.6 before the next major version is published, they are more than welcome to submit a PR with the updated GH workflow tests to add back support for 3.6.

_Originally posted by @samj1912 in https://github.com/PyCQA/pydocstyle/issues/539#issuecomment-842563433_

Python 3.6 support was removed in the above PR. Creating this issue to track if adding back support for 3.6 is feasible before the next major release of pydocstyle.

RuRo:
This reverts commit f710d32431fa5662e8c08c305d699e506c240f0b.
This closes #540.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 7.0.0rc
+current_version = 6.1.1rc
 commit = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)((?P<release>.*))?
 serialize = 
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -8,10 +8,6 @@ Release Notes
 Current Development Version
 ---------------------------
 
-Major Updates
-
-* Support for Python 3.6 has been dropped (#539).
-
 
 6.1.0 - May 17th, 2021
 ---------------------------
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 # Do not update the version manually - it is managed by `bumpversion`.
-version = '7.0.0rc'
+version = '6.1.1rc'
 
 
 requirements = [
@@ -26,6 +26,7 @@ setup(
         'Environment :: Console',
         'Development Status :: 5 - Production/Stable',
         'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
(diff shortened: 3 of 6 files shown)


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 7.0.0rc
+current_version = 6.1.0
```
samj1912, line 2:
```suggestion
current_version = 6.1.1rc
```

In case this passes

setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 # Do not update the version manually - it is managed by `bumpversion`.
-version = '7.0.0rc'
+version = '6.1.0'
```
samj1912, line 4:
```suggestion
version = '6.1.1rc'
```

setup.py
```diff
@@ -26,14 +26,14 @@
         'Environment :: Console',
         'Development Status :: 5 - Production/Stable',
         'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
-        'Programming Language :: Python :: 3.9',
```
samj1912, line 31: This should still be around

src/pydocstyle/utils.py
```diff
@@ -6,7 +6,7 @@
 from typing import Any, Iterable, Tuple
 
 # Do not update the version manually - it is managed by `bumpversion`.
-__version__ = '7.0.0rc'
+__version__ = '6.1.0'
```
samj1912, line 9:
```suggestion
__version__ = '6.1.1rc'
```

docs/release_notes.rst
```diff
@@ -5,14 +5,6 @@ Release Notes
 `Semantic Versioning <http://semver.org/>`_ specification.
 
 
-Current Development Version
----------------------------
```
samj1912, line 9: We should probably leave this around.

samj1912:
One thing that irks me is why the CI suddenly started failing. I am guessing it is because of the new version of click that was released. Instead of changing the locale, you may want to limit the click version to click<8.0 in the dev requirements instead.

RuRo:
I think, the locale setting update worked. I'd rather fix the CI directly than pinning dependency versions as it doesn't look like the new click version **actually** broke any behaviour. But I can also pin the version if you insist.

docs/release_notes.rst
RuRo: Done

samj1912:
Since this is just used by black (for development only) I am fine either way.

samj1912 (approved):

Issue #531, opened 2021-04-09
D401 "not in imperative mood" not truly applicable to @property


ragazzojp:
`@property` methods are not used as methods. Their docstrings are used to set the description of the properties. Errors like D401 "docstring not in imperative mood" should not be applicable to methods, at least to getters, that define a `@property`.

IMHO, the following docstring should be accepted for `x`, although not in imperative mood:

```python
@property
def x(self):
    "The X property"
    return self._x
```

Alternatively, please classify these errors differently, so users can switch them off.

ragazzojp:
Same for DAR201, I don't think `:returns:` is mandatory when the property description is a multiline string.

samj1912:
You should be able to use `--ignore-decorators` to ignore pydocstyle errors for certain decorated callables.

http://www.pydocstyle.org/en/stable/usage.html

`DAR201` is not raised by pydocstyle.

WasabiFan:
The above suggested workaround (`--ignore-decorators`) seems to only allow entirely disabling pydocstyle for all `@property` docstrings. I appreciate the availability of a workaround but I think this one isn't sufficiently surgical to effectively unblock a user here.

What _is_ the suggested/intended "mood" for a property docstring otherwise? I guess "Return the X property"?

sigmavirus24:
> but I think this one isn't sufficiently surgical to effectively unblock a user here.

What is blocking a user from your perspective? How should this be more "surgical"?

> What _is_ the suggested/intended "mood" for a property docstring otherwise? I guess "Return the X property"?

How do you document the attributes/properties of classes otherwise?

samj1912:
Fixed by #546

PR #546, opened 2021-06-17
Exempt properties from D401

TomFryers:
This addresses issue #531. It excludes any method decorated with a decorator with "property" in the name in order to catch `cached_property` and other custom property types. Maybe there’s a fancier way of doing this, but I don’t know what it is.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -8,9 +8,9 @@ Release Notes
 Current Development Version
 ---------------------------
 
-Bug Fixes
+New Features
 
-* Properties no longer cause D401
+* Add support for `property_decorators` config to ignore  D401
 
 6.1.1 - May 17th, 2021
 ---------------------------
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -8,6 +8,9 @@ Release Notes
 Current Development Version
 ---------------------------
 
+Bug Fixes
+
+* Properties no longer cause D401
 
 6.1.1 - May 17th, 2021
 ---------------------------
--- a/src/pydocstyle/parser.py
+++ b/src/pydocstyle/parser.py
@@ -219,6 +219,14 @@ class Function(Definition):
         return False
 
     @property
+    def is_property(self):
+        """Return True iff the method decorated with property."""
+        for decorator in self.decorators:
+            if "property" in decorator.name:
+                return True
+        return False
+
+    @property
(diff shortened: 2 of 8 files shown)


src/pydocstyle/parser.py
```diff
@@ -218,6 +218,14 @@ def is_overload(self):
                 return True
         return False
 
+    @property
+    def is_property(self):
+        """Return True iff the method decorated with property."""
+        for decorator in self.decorators:
+            if "property" in decorator.name:
```
sigmavirus24, line 225: Pretty sure if I create a decorator named `myproperty` this will disable D401 which is far from ideal. Why not `if decorator.name == "property":`?

src/pydocstyle/parser.py
```diff
@@ -218,6 +218,14 @@ def is_overload(self):
                 return True
         return False
 
+    @property
+    def is_property(self):
+        """Return True iff the method decorated with property."""
```
sigmavirus24, line 223: This says `iff` which tends to only mean "if and only if" but that's not actually true here because anything with the string "property" in the name will return True, not just the `@property` built-in decorator

sigmavirus24:
So if you want to catch both `@property` and `@cached_property` then I'd suggest the best way is to add a flag called `--property-function-names` (or something like that) and then use that to do `if decorator.name in options.property_fn_names`

src/pydocstyle/parser.py
```diff
@@ -218,6 +218,13 @@ def is_overload(self):
                 return True
         return False
 
+    def is_property(self, decorators):
+        """Return True iff the method decorated with property."""
+        for decorator in self.decorators:
```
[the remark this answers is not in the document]
TomFryers: I’ve improved this. I also changed the above `is_overload` to match.

docs/release_notes.rst
```diff
@@ -8,6 +8,9 @@ Release Notes
 Current Development Version
 ---------------------------
 
+Bug Fixes
+
+* Properties no longer cause D401
```
samj1912, line 13:
```suggestion
New Features

* Add support for `property_decorators` config to ignore  D401
```

samj1912 (approved):

samj1912 (approved):

samj1912:
@TomFryers sorry for the delay. Thank you very much for this contribution :)!

PR #529, opened 2021-04-01
Add a match_path option for comparison against a full path.

heoga:
It has been noted in issue #363 that match & match_dir are
unwieldy when attempting to match against full paths.  For
unexample if you have A.py in directories B & C and you only
want to run pydocstyle on one of them.

From my own experience trying to deploy pydocstyle against a
large legacy codebase it is unworkable as it would mean the
entire codebase being converted as a big bang change.

This commit adds a new option, match_path, to the config &
command lines which can be used to provide more nuanced
matching.  For example the following specification:

    match_path = [AB]/[ab].py
                 D/e.py

This defines two regexes.  If either match a given path,
relative to the directory specified, the file will be yielded
for comparison.  The is complimentary to match & match_dir and
the three can be used together.

A test has been added to test_integration to test & illustrate the feature.
tox was run on Linux (Ubuntu under the subsystem for Windows) and all tests were successful.

PR #550, opened 2021-08-16
Fix match option to only consider basename when given a path argument

### LINKED ISSUE #549: --match should consider only basename when given a path argument
As an example,

```
pydocstyle --match '(?!setup).*\.py' setup.py
```

correctly ignores `setup.py`. However, both

```
pydocstyle --match '(?!setup).*\.py' .\setup.py
```

and

```
pydocstyle --match '(?!setup).*\.py' ./setup.py
```

don't. This behavior is confusing for a user. Even worse, IDEs like VSCode automatially invoke `pydocstyle` with full path of opened Python file. Even if I include a `match` pattern in my configuration, it doesn't work because full paths are passed to `pydocstyle`. I wasted some time trying to figure out if I had configured `pydocstyle` correctly because VSCode was still flagging problems in ignored files.

I think the issue can be resolved by only considering basename of file arguments to match against `--match` pattern.

oczkoisse:
Fixes #549.

This uses `os.path.basename` before trying to match against the pattern set in `--match` option. This way, even if full paths to files are passed to `pydocstyle` (by VSCode, for example), only filename part of the path will be matched with the set pattern, thereby eliminating accidental matches with the path. This is also more consistent with how `--match` is documented previously.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -12,6 +12,10 @@ New Features
 
 * Add support for `property_decorators` config to ignore  D401
 
+Bug Fixes
+
+* Fix ``--match`` option to only consider filename when matching full paths (#550).
+
 6.1.1 - May 17th, 2021
 ---------------------------
 
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -288,7 +288,7 @@ class ConfigurationParser:
                     # Skip any dirs that do not match match_dir
                     dirs[:] = [d for d in dirs if match_dir(d)]
 
-                    for filename in filenames:
+                    for filename in map(os.path.basename, filenames):
                         if match(filename):
                             full_path = os.path.join(root, filename)
                             yield (
@@ -302,7 +302,7 @@ class ConfigurationParser:
                 match, _ = _get_matches(config)
                 ignore_decorators = _get_ignore_decorators(config)
                 property_decorators = _get_property_decorators(config)
-                if match(name):
+                if match(os.path.basename(name)):
                     yield (
                         name,
                         list(config.checked_codes),
--- a/src/tests/test_integration.py
+++ b/src/tests/test_integration.py
@@ -1489,3 +1489,25 @@ def test_comment_with_noqa_plus_docstring_file(env):
     out, _, code = env.invoke()
     assert '' == out
     assert code == 0
+
+
+def test_match_considers_basenames_for_path_args(env):
+    """Test that `match` option only considers basenames for path arguments.
+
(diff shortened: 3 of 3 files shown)


src/pydocstyle/config.py
```diff
@@ -288,7 +288,7 @@ def _get_property_decorators(conf):
                     # Skip any dirs that do not match match_dir
                     dirs[:] = [d for d in dirs if match_dir(d)]
 
-                    for filename in filenames:
+                    for filename in map(os.path.basename, filenames):
                         if match(filename):
```
samj1912, line 292:
```suggestion
                    for filename in filenames:
                        if match(filename) or match(os.path.basename(filename)):
```

src/pydocstyle/config.py
```diff
@@ -302,7 +302,7 @@ def _get_property_decorators(conf):
                 match, _ = _get_matches(config)
                 ignore_decorators = _get_ignore_decorators(config)
                 property_decorators = _get_property_decorators(config)
-                if match(name):
+                if match(os.path.basename(name)):
```
samj1912, line 305:
```suggestion
                if match(name) or match(os.path.basename(name)):
```

docs/release_notes.rst
```diff
@@ -15,6 +15,10 @@ New Features
 * Add support for Python 3.10 (#554).
 * Replace D10X errors with D419 if docstring exists but is empty (#559).
 
+Bug Fixes
+
+* Fix ``--match`` option to only consider filename when matching full paths (#550).
```
samj1912, line 20:
```suggestion
* Fix ``--match`` option to consider both the base filename and the full-path when matching full paths (#550).
```

samj1912 (approved):

samj1912:
@oczkoisse 👋 Thank you so much for the fix and sorry for the delay in reviewing the PR. After my latest commit [`e5880d5` (#550)](https://github.com/PyCQA/pydocstyle/pull/550/commits/e5880d57b163c7b6c82de0fb3d43de9b4fd9f27b) in order to preserve b/w compat for match patterns like (`src\/*\.py`) i.e patterns that match a full path for eg. - it seems like it breaks the test you added for the negative regex matches.

I am in a difficult conundrum with this since one way breaks b/w compat for nested match paths and the other causes issues with negative match directives. 

Re-thinking this PR in the context of #529 - I think it might be better for me to revert my latest commit and merge #529 to have proper support for nested dir paths.

oczkoisse:
Thanks @samj1912 for taking the time to review this PR. I think #529 is a good way to provide an alternative to people who rely on `--match` to match against nested paths.

PR #556, opened 2021-11-10
Sections

### LINKED ISSUE #463: Spurious D401 when docstring only contains a Returns section
With
```python
class T:
    def get_x(self):
        """
        Returns
        -------
        int
            The x.
        """
        return self.x
```
pydocstyle currently warns (in particular)
```
/tmp/test.py:3 in public method `get_x`:
        D205: 1 blank line required between summary line and description (found 0)
/tmp/test.py:3 in public method `get_x`:
        D400: First line should end with a period (not 's')
/tmp/test.py:3 in public method `get_x`:
        D401: First line should be in imperative mood (perhaps 'Return', not 'Returns')
```
The first two warnings are reasonable (they are effectively saying "you are missing a summary line" -- a single warning for that would be nicer but things are fine as is), but the last one is spurious: one should be able to disable D205/D400 to allow docstrings without a single summary line while keeping the check for imperative mood *when a summary line is present*.  (pydocstyle already checks for sections in the docstring, so I guess it could just disable D401 when it detects that the first line is indeed a section title?)

anntzer:
Skip D401 for docstrings that start directly with a section.  Closes #463.


anntzer:
Kindly bumping.

PR #560, opened 2021-11-22
Add `ignore-self-only-init` option

thejcannon:
Adding an option `ignore-self-only-init` (off by default) which will not report a D107 error if `__init__` has no parameters other than `self`. The justification is that these methods usually aren't very interesting and will likely result in redundant/boring docstrings which add noise.

Just turning off D107 isn't an option, since there are `__init__` methods which _are_ interesting ;)


thejcannon:
Of course, the option name is up for debate :)

src/pydocstyle/parser.py
```diff
@@ -665,8 +671,10 @@ def parse_definition(self, class_):
         name = self.current.value
         self.log.debug("parsing %s '%s'", class_.__name__, name)
         self.stream.move()
+        call_names = []
```
samj1912, line 674: Should we name this `callable_args` instead?

samj1912 (commented):
Hey @thejcannon - thanks for the PR! Sorry for the delay in getting to review this. It generally looks good to me. Just a minor comment re:naming.

src/pydocstyle/parser.py
thejcannon: At this point I forgot why I called it this :joy:

`callable_args` it is!

samj1912:
@thejcannon could you please rebase this PR against the master branch? Thanks!

thejcannon:
@samj1912 got time for a quick review?

docs/release_notes.rst
```diff
@@ -35,6 +35,7 @@ New Features
 * Allow for hanging indent when documenting args in Google style. (#449)
 * Add support for `property_decorators` config to ignore D401.
 * Add support for Python 3.10 (#554).
+* Add `ignore-self-only-init` config (#560).
```
samj1912, line 38: Can we move this to the appropriate version section?

samj1912 (approved):
Minor comment about release notes. Lgtm apart from that.

docs/release_notes.rst
thejcannon: done

PR #559, opened 2021-11-22
Add D419: Add and switch to "Docstring is empty" error code

thejcannon:
Reporting D10X errors for empty docstrings is doubly misleading:
1. The docstring _isn't_ missing. It's _empty_.
2. The D10X errors use _public_, but could be reported on private definitions.

Please make sure to check for the following items:
- [x] Add unit tests and integration tests where applicable.  
- [x] Add a line to the release notes (docs/release_notes.rst) under "Current Development Version".

PR #561, opened 2021-11-23
Change appropriate "Nested" definitions into "Inaccessible", and add new codes

thejcannon:
Some devs want to be able to ignore inaccessible definitions, and some might want to not. This should allow them the option to choose.

(This also helps keep the definition of `is_public` pure and straightforward and correct according to the documentation)


src/tests/test_cases/test.py
```diff
@@ -67,10 +67,11 @@ def __call__(self=None, x=None, y=None, z=None):
 @expect('D103: Missing docstring in public function')
 def function():
     """ """
+    @expect('D123: Missing docstring in inaccessible public function')
     def ok_since_nested():
         pass
 
-    @expect('D103: Missing docstring in public function')
+    @expect('D123: Missing docstring in inaccessible public function')
```
thejcannon, line 74: This change should be superseded by https://github.com/PyCQA/pydocstyle/pull/559

src/pydocstyle/config.py
```diff
@@ -183,6 +183,7 @@ class ConfigurationParser:
     )
     BASE_ERROR_SELECTION_OPTIONS = ('ignore', 'select', 'convention')
 
+    DEFAULT_IGNORE = {"D121", "D123"}
```
thejcannon, line 186: Other tools have this kind of "we have the ability to report these codes, but their off by default" functionality. I figured this belonged to that category. Note if people are selecting `ignore`s today, this will no longer be the default.

thejcannon (commented):
I couldn't find a good term after scouring python.org for what this might be called. I wanted to avoid "nested" because that exists for classes. "Inner" also didn't feel correct because methods are _inside_ classes. I landed on "inaccessible" because the caller can't access them.

src/pydocstyle/parser.py
```diff
@@ -17,10 +17,10 @@
     'Module',
     'Package',
     'Function',
-    'NestedFunction',
-    'Method',
+    'InaccessibleFunction' 'Method',
```
thejcannon, line 20:
```suggestion
    'InaccessibleFunction', 
    'Method',
```

thejcannon, line 20:
```suggestion
    'InaccessibleFunction', 
    'Method',
```

src/pydocstyle/parser.py
```diff
@@ -17,10 +17,11 @@
     'Module',
     'Package',
     'Function',
-    'NestedFunction',
+    'InaccessibleFunction',
```
thejcannon, line 20:
```suggestion
    'InaccessibleFunction',
```

thejcannon:
@samj1912 this should be ready to review :smile:

thejcannon:
@samj1912 got time for a quick review?

PR #570, opened 2022-01-10
Fix TypeError with bytes docstrings

### LINKED ISSUE #569: Crash: TypeError: a bytes-like object is required, not 'str'
Code snippet:

```python
def f():
    b"""abc"""
```

Output:
```console
$ pydocstyle mytest.py 
  File "/Users/tusharsadhwani/code/marvin-python/venv3/bin/pydocstyle", line 8, in <module>
    sys.exit(main())
  File "/Users/tusharsadhwani/code/marvin-python/venv3/lib/python3.10/site-packages/pydocstyle/cli.py", line 68, in main
    sys.exit(run_pydocstyle())
  File "/Users/tusharsadhwani/code/marvin-python/venv3/lib/python3.10/site-packages/pydocstyle/cli.py", line 44, in run_pydocstyle
    errors.extend(check((filename,), select=checked_codes,
  File "/Users/tusharsadhwani/code/marvin-python/venv3/lib/python3.10/site-packages/pydocstyle/checker.py", line 993, in check
    for error in ConventionChecker().check_source(source, filename,
  File "/Users/tusharsadhwani/code/marvin-python/venv3/lib/python3.10/site-packages/pydocstyle/checker.py", line 143, in check_source
    error = this_check(self, definition,
  File "/Users/tusharsadhwani/code/marvin-python/venv3/lib/python3.10/site-packages/pydocstyle/checker.py", line 204, in check_one_liners
    lines = safe_literal_eval(docstring).split('\n')
TypeError: a bytes-like object is required, not 'str'
```

tushar-deepsource:
Let me know if the fix is adequate or if it should be fixed in some other way.

Resolves #569 

Pending:

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -246,7 +246,12 @@ class ConventionChecker:
 
         """
         if docstring:
-            lines = ast.literal_eval(docstring).split('\n')
+            docstring_text = ast.literal_eval(docstring)
+            if isinstance(docstring_text, bytes):
+                docstring_text = docstring_text.decode()
+
+            lines = docstring_text.split('\n')
+
             if len(lines) > 1:
                 non_empty_lines = sum(1 for l in lines if not is_blank(l))
                 if non_empty_lines == 1:


Pierre-Sassoulas:
Seems like I can't approve the pipeline, do you know what's wrong with the rights @samj1912 ?

sigmavirus24 (changes requested):
There should be an error here otherwise we're silently ignoring bytes docstrings and that is just as wrong as crashing

tusharsadhwani:
@sigmavirus24 are you suggesting there be another lint rule?

sigmavirus24:
I'm suggesting that skipping byte string docstrings is going to be confusing for users. So we either need to try to do the right thing or emit an error code that this is a byte string and those aren't valid docstrings (because they're not if I understand the old discussion properly).

tusharsadhwani:
@sigmavirus24 I'm in favour of simply raising a new issue for bytes then.

Trying to do the right thing has its own edge case: if we're unable to decode the bytes object, we can't do much.

Issue #555, opened 2021-10-17
D102 triggered for functions decorated with `typing.overload` rather than `overload`


sigmavirus24:
Related to #419 but not quite

As a result of #419, pydocstyle now checks the decorators of a function such that if they match the name `overload` ([and only that](https://github.com/PyCQA/pydocstyle/blob/d4d9d6342ff201b5bb4b672ae35ac3e43968190a/src/pydocstyle/parser.py#L214)) then it skips checking for docstrings. However, not all people like to import from `typing` and instead import it or import alias it, e.g., 

```py
import typing
import typing as t
```

This means that then doing

```py
@t.overload
def foo(s: str) -> str:
    ...

# alternatively
@typing.overload
def foo(s: str) -> str:
    ...
```

Triggers a D102.

PR #571, opened 2022-01-13
Fix issues with `typing.overload`

### LINKED ISSUE #525: Wrong "D105: Missing docstring in magic method" for overload function
Similar to #419 but for class magic methods.

The following example results in the the two overloaded `__getitem__` methods to be flagged with D105. The expected result is that they are not flagged.

```python
from typing import Iterable, List, MutableSequence, TypeVar, overload

_T = TypeVar("_T")

class MyClass(MutableSequence(_T)):
    """Example class."""

    def __init__(self, iterable: Iterable[_T]) -> None:
        """Instantiate class."""
        self._data = list(iterable)

    @overload
    def __getitem__(self, __index: int) -> _T:
        ...

    @overload
    def __getitem__(self, __index: slice) -> List[_T]:
        ...

    def __getitem__(self, __index):
        """Get item by index."""
        return self._data[__index]
```

yumasheta:
This should fix #525 and also the `def __init__()` overloading which is not explicitly part of that issue.
The fix simply makes use of the existing `is_overload` property also in the case that `definition.is_magic` is true. Previously, this was somehow missed in the inlined if-else construct. I expanded that part to more explicit statements and used a new variable.

I also tried #555 and at least added a match for "typing.overload". But I don't know if pydocstyle is able to infer if a decorator is actually a name bound at import by something like: `from typing import overload as my_overload`.
If not, then I don't see a way to "know" if a given decorator is a `typing.overload`.

If this is ok, I'd add tests as well.

TODO:

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -200,21 +200,20 @@ class ConventionChecker:
 
         """
         if not docstring and definition.is_public:
+            method_violations = None
+            if not definition.is_overload:
+                if definition.is_magic:
+                    method_violations = violations.D105()
+                elif definition.is_init:
+                    method_violations = violations.D107()
+                else:
+                    method_violations = violations.D102()
+
             codes = {
                 Module: violations.D100,
                 Class: violations.D101,
                 NestedClass: violations.D106,
-                Method: lambda: violations.D105()
-                if definition.is_magic
-                else (
-                    violations.D107()
-                    if definition.is_init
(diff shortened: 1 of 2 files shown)


yumasheta:
I noticed a strange side effect. When using the changes of this PR, D105 is never detected.

yumasheta:
Should be fixed now. I missed that the evaluation of the error `codes` dict is delayed.

PR #577, opened 2022-02-16
Fix decorator parsing for async functions

cdce8p:
This PR addresses an issue with parsing decorators right before async function definitions.
A good example to illustrate the issue is `overload`.
```py
from typing import overload

@overload
async def func(x: int) -> int:
    ...

async def func(x: int | str) -> int | str:
    """Do something."""
    pass
```
At the moment, this would emit a D103 error for the first definition.
```
```

That's because, parsing the decorator name wasn't stopped with `async` and thus `overload` became `overloadasync` and didn't match any `is_overload` checks anymore.


docs/release_notes.rst
```diff
@@ -25,6 +25,8 @@ New Features
 Bug Fixes
 
 * Fix ``--match`` option to only consider filename when matching full paths (#550).
+* Fix decorator parsing for async function. Resolves some false positives
```
samj1912, line 28: @cdce8p thanks for the PR! Could you please move the release notes under a 6.2.1 heading instead? Thanks!

cdce8p:
Moved the changelog entry again, since I didn't saw that `6.2.1` was released an hour ago.

Pierre-Sassoulas (approved):

samj1912 (approved):

Issue #575, opened 2022-01-22
Call for maintainers 


samj1912:
Hello all,

As of late, I have been the only somewhat active maintainer for pydocstyle and I am in a position where I am finding less and less time to dedicate to this project.

As a consequence, @Nurdok and I have decided to do a call for maintainers.

Ideally we would love for the new maintainer to

- be an avid user or contributor to pydocstyle or a similar project (like sibling linting projects in PyCQA)
-  have a good history of involvement in open source 

The immediate tasks for new maintainer(s) would be to - 

- Automate the release process 
- Release the current set of contributions on the master branch
- Triage the issue backlog
- Review the current set of open PRs 

Longer term, there are a few milestones on - 

- Making conventions a first class concepts in pydocstyle as opposed to a group of issues 
- Enhance the linting capabilities for numpy and Google docstyles 

for more details see https://github.com/PyCQA/pydocstyle/milestone/7

If you are interested, feel free to reach out to us on the pydocstyle #general discord channel https://discord.gg/GAt2YQEXga or comment on this issue.

Pierre-Sassoulas:
Hello @samj1912, 

First, did you consider Tidelift ? [The current estimated income is 15$/month](https://tidelift.com/lifter/search/pypi/pydocstyle), it's not much, but it could justify spending an hour on it per month. 

Second I'm maintaining pylint, and pylint is not using pydocstyle directly but often have check that are very close or even duplicate (we handle numpy/google docstyle in our docstring plugin). Knowing pydocstyle better could be a way to rationalize pylint and maybe use pydocstyle directly inside it. Maybe a way to decrease the maintenance required for both.

I'm not saying I'd have the time to triage everything and review all the PR but I can at least automate the release process and release periodically because I already did that for pylint/astroid/prospector so it's... probably not much than copy pasting yaml in ``.git`` and asking you for the proper rights on pypi or defining an env variable in the github settings.

Let me know what you think :)

samj1912:
@Pierre-Sassoulas that sounds great! I am away on vacations for the next few weeks so I might be slow to respond but if you send over a github actions PR, I can make sure proper secrets are injected in the workflows. 

I have not considered Tidelift. Thank you for the link, I will take a look :) I don't think that will massively change much, with regards to the current time I have available for pydocstyle.

samj1912:
[...]

PR #582, opened 2022-03-10
Add github action to be able to release on published tag

Pierre-Sassoulas:
This permit to release on pypi if a value for ``PYPI_API_TOKEN`` is set in github settings. The release is done when a release is created/published in github interface.

Relates to #575

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,37 @@
+name: Release
+
+on:
+  release:
+    types:
+      - published
+
+env:
+  DEFAULT_PYTHON: 3.9
+
+jobs:
+  release-pypi:
+    name: Upload release to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code from Github
+        uses: actions/checkout@v3.0.0
+      - name: Set up Python ${{ env.DEFAULT_PYTHON }}
+        id: python
+        uses: actions/setup-python@v3.0.0
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON }}
+      - name: Install requirements
+        run: |
+          python -m pip install -U pip twine wheel
+          python -m pip install -U "setuptools>=56.0.0"
+      - name: Build distributions
+        run: |
+          python setup.py sdist bdist_wheel
+      - name: Upload to PyPI
+        if: github.event_name == 'release' && startsWith(github.ref, 'refs/tags')
+        env:
+          TWINE_REPOSITORY: pypi
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
+        run: |
+          twine upload --verbose dist/*
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,9 +16,6 @@ jobs:
     name: Upload release to PyPI
     runs-on: ubuntu-latest
     environment:
-      name: pypi-dev
-      url: https://test.pypi.org/project/pydocstyle/
-    environment:
       name: pypi-prod
       url: https://pypi.org/project/pydocstyle/
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,11 +29,10 @@ jobs:
           check-latest: true
       - name: Install requirements
         run: |
-          python -m pip install -U pip twine wheel
-          python -m pip install -U "setuptools>=56.0.0"
+          python -m pip install twine build
       - name: Build distributions
         run: |
-          python setup.py sdist bdist_wheel
+          python -m build
       - name: Upload to PyPI
         if: github.event_name == 'release' && startsWith(github.ref, 'refs/tags')
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,8 @@ jobs:
   release-pypi:
     name: Upload release to PyPI
     runs-on: ubuntu-latest
+    environment:
+      name: pypi-prod
     steps:
       - name: Check out code from Github
         uses: actions/checkout@v3.0.0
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This should be a protected workflow so only trusted folks can do this, no?

Pierre-Sassoulas:
> This should be a protected workflow so only trusted folks can do this, no?

The workflow is triggered by [creating a release](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) (in github interface from this url https://github.com/PyCQA/pydocstyle/releases) so only members that can release a package can trigger this workflow.

sigmavirus24:
I wasn't sufficiently clear. There are often cases where you don't want everyone with permission to create a release to be able to publish to PyPI. No one can pip install from GitHub and get malicious code (although some people apparently do install artifacts from there which boggles my mind). Publishing to PyPI should be more severely restricted. I thought @sethmlarson had written a blog post about this but I'm not seeing it.

Pierre-Sassoulas:
I don't see the purpose of a release without a pypi publication alongside it, so for me those permissions were equivalent. Let me know if I need to modify something :)

> some people apparently do install artifacts from there which boggles my mind

I would have made the mistake 15 years ago. It's probably autodidacts with no formal training who were never introduced to package management. When you come from a Windows background you're used to searching for an executable to install after searching for the name of the software online, and github releases are what's closest to this.

sethmlarson:
@sigmavirus24 Article you're looking for is here: https://sethmlarson.dev/blog/security-for-package-maintainers

sigmavirus24:
> for me those permissions were equivalent

They shouldn't always be equivalent. That's my point. Also, GitHub releases are trash for so many reasons and create extra burden on maintainers. They shouldn't be the vehicle to publishing to PyPI

Pierre-Sassoulas:
It's also possible to launch a github actions on tag creation directly but there would still be the problem of who has right to create tag. If you don't want to mix pypi / github rights then the solution is probably a local script launched by someone with pypi's release right.

samj1912:
@Pierre-Sassoulas - I have created two environments `pypi-dev` (which has tokens for test.pypi) and `pypi-prod` (which has tokens for pypi) both of which are gated and require approvers from the current admins (me and @Nurdok) who have release permissions on PyPi. Could you follow something similar to https://timheuer.com/blog/add-approval-workflow-to-github-actions/ to add these environments to the workflow? 

I have created project specific tokens and added them as env vars to these environments.

samj1912:
Separately - it would have been great if we could have versions also driven by the Github tags instead having to manually bump them similar to https://github.com/cruft/cruft/blob/0c6de85d974197969c0e65913857eaa36b788e5e/.github/workflows/pypi_publish.yml#L22

but any improvement to the release process is great. Thanks for this PR! 🙌

Pierre-Sassoulas:
@samj1912 I don't know if I understood your first link correctly, Would [342cc17](https://github.com/PyCQA/pydocstyle/pull/582/commits/342cc1780f60ca030d39d906b7bfc47f396690a6) limit the person that can launch the job ? In order to test we need to create a tag and publish a release which is... something that I'd try on another smaller repo, personally 😄 

Second link look nice too but I have very limited available time and I'm not knowledgeable enough about poetry right now.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON }}
+          check-latest: true
+      - name: Install requirements
+        run: |
+          python -m pip install -U pip twine wheel
+          python -m pip install -U "setuptools>=56.0.0"
+      - name: Build distributions
+        run: |
+          python setup.py sdist bdist_wheel
```
adamjstewart, line 39: Direct invocation of `setup.py` is [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html), current [best practice](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives) is to use `python -m build` instead.

adamjstewart (commented):
Most of the projects I maintain use [pypa/cibuildwheel](https://github.com/pypa/cibuildwheel) and [actions/upload-artifact](https://github.com/actions/upload-artifact) instead of manual `build` and `twine`. May be more reliable since they are maintained by PyPA or GitHub themselves.

Pierre-Sassoulas:
> Most of the projects I maintain use [pypa/cibuildwheel](https://github.com/pypa/cibuildwheel) and [actions/upload-artifact](https://github.com/actions/upload-artifact) instead of manual build and twine. May be more reliable since they are maintained by PyPA or GitHub themselves.

This looks pretty cool, are wheel properly created for al operating systems with ``python -m build`` ? Because otherwise we should definitely use what you suggested. (I'm using this release workflow in a number of places so I appreciate the review, thank you @adamjstewart  :) )

adamjstewart:
> are wheel properly created for al operating systems with `python -m build`?

In the last couple of years, PyPA completely overhauled their support for non-default build backends. Build "backends" include things like distutils, setuptools, flit, hatchling, poetry, etc. Build "frontends" include things like build and pip. One of the advantages of build and pip is that they support all of the above build backends, whereas setup.py only supports distutils/setuptools. Also setup.py is deprecated as I mentioned.

`python -m build` basically does the same thing that `python setup.py build` used to do. For pure-Python projects like pydocstyle, it creates a single universal wheel for Python 3. For Python libraries that depend on external libraries or that require compilation, it builds a Python-version and platform-specific wheel, and you need to run this command on all versions/platforms (e.g., via matrix in GitHub Actions) and upload each one separately. 

cibuildwheel is probably overkill for this project since we don't need a separate wheel for each Python version/platform. Let's stick with build/twine instead.

Pierre-Sassoulas:
Thanks a lot for the explanation, I only have pure python project right now but I'll keep that in mind 👍

samj1912 (approved):

adamjstewart (approved):

samj1912:
Thank you so much for this!

PR #584, opened 2022-03-13
Fix test warnings and improve lint configuration

aphedges:
Thanks for submitting a PR!

---

I ran the tests locally when I noticed some pytest warnings:

```python
.tox/py36-tests/lib/python3.6/site-packages/_pytest/config/__init__.py:1233
  /Users/ahedges/projects/personal/pydocstyle/.tox/py36-tests/lib/python3.6/site-packages/_pytest/config/__init__.py:1233: PytestConfigWarning: Unknown config option: pep8ignore
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

src/tests/test_integration.py:124
  /Users/ahedges/projects/personal/pydocstyle/src/tests/test_integration.py:124: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    @pytest.yield_fixture(scope="module")

src/tests/test_integration.py:141
  /Users/ahedges/projects/personal/pydocstyle/src/tests/test_integration.py:141: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    @pytest.yield_fixture(scope="function", params=['ini', 'toml'])
```

I also saw a Sphinx warning: `WARNING: html_static_path entry '/Users/ahedges/projects/personal/pydocstyle/docs/_static' does not exist`.

I decided to fix these warnings.

I can add release notes if you want, but I don't think they are needed for this.

## Changes

- Fix pytest deprecation warnings by replacing `@pytest.yield_fixture` with `@pytest.fixture`
- Fix pytest unused configuration option warning by removing `pep8ignore`
  - It was used by `pytest-pep8`, which was removed from testing in 0a78a1df7462818e8d405d5f5752e0402baa05c1.
- Fix Sphinx warning by commenting out a line similar to nearby options
  - There hasn't been a `_static` directory since the the line was first added in e8fd68a9cc544aca2040d08ca08edd29e0bd93a4.
- Move mypy config from `tox.ini` to `pyproject.toml`
  - While this isn't strictly necessary, `tox.ini` is not one of the places mypy looks for configuration by default, so running mypy by itself was more annoying. The pinned version of mypy supports configuration in `pyproject.toml`, so it made sense to move it in with the Black and isort configuration.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -66,8 +66,6 @@ skip_install = {[testenv:install]skip_install}
 commands = {[testenv:install]commands}
 
 [pytest]
-pep8ignore =
-    test.py E701 E704
 norecursedirs = docs .tox
 addopts = -rw
 
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,7 @@ setenv =
 # To pass arguments to pytest, use `tox [options] -- [pytest posargs]`.
 commands =
     pytest --cache-clear -vv src/tests {posargs}
-    mypy --config-file=tox.ini src/
+    mypy src/
     black --check src/pydocstyle
     isort --check src/pydocstyle
 deps =
@@ -73,8 +73,3 @@ addopts = -rw
 inherit = false
 convention = pep257
 add-select = D404
-
-[mypy]
-ignore_missing_imports = true
-strict_optional = true
-disallow_incomplete_defs = true
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -134,7 +134,7 @@ else:
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+#html_static_path = ['_static']
 
 # Add any extra paths that contain custom files (such as robots.txt or
 # .htaccess) here, relative to this directory. These files are copied
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,3 +7,8 @@ skip-string-normalization = true
 profile = "black"
 src_paths = ["src/pydocstyle"]
 line_length = 79
+
+[tool.mypy]
+ignore_missing_imports = true
+strict_optional = true
+disallow_incomplete_defs = true
--- a/src/tests/test_integration.py
+++ b/src/tests/test_integration.py
@@ -121,7 +121,7 @@ class SandboxEnv:
         pass
 
 
-@pytest.yield_fixture(scope="module")
+@pytest.fixture(scope="module")
 def install_package(request):
     """Install the package in development mode for the tests.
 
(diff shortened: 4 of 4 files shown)


Mr-Pepe (approved):
Looks good to me

tox.ini
```diff
@@ -16,7 +16,7 @@ setenv =
 # To pass arguments to pytest, use `tox [options] -- [pytest posargs]`.
 commands =
     pytest --cache-clear -vv src/tests {posargs}
-    mypy --config-file=tox.ini src/
```
sigmavirus24, line 19: What test warning does this fix?

Mr-Pepe: Good point, contrary to the title this does not fix a warning. However it makes the usage of mypy less verbose outside of tox which I think is a good thing. Maybe this should be part of a separate PR that improves tool configurations? I could open a PR that moves all tool-specific configuration to the pyproject.toml and also changes the deprecated [pep257] section name to [pydocstyle].

sigmavirus24: If the point is to simplify usage for people refusing to use the project's tooling, we should move `pytest` options here to `add-opts` below rather than here, and we should have some other way to run `black` and `isort` because those are too cumbersome too, clearly, right?

People using pycharm + pytest won't have `--clear-cache` automatically configured for them or `-vv`. 

The point of using `tox` for things like this is to simplify things.

Mr-Pepe: I often just run a single tool from the tox pipeline (e.g., mypy while fixing type hints) and find it cumbersome to look up the specific command line options in the tox.ini. I therefore put all configuration in configuration files and only have calls like `pytest src/tests` or `mypy src/` in the tox.ini. So yes, I think moving `--clear-cache` and `-vv` into the `add-opts` section would be great because it centralizes the configuration. Am I missing some drawback of doing so? If so, why is pytest's `-rw` option in `add-opts`?

What other way to call black and isort do you mean? Their configs are already in the pyproject.toml so I can simply run `black src/pydocstyle` without worrying about additional command line options.

I am not sure about Pycharm but setting up tools like black and isort with VS Code picks up the configuration from pyproject.toml.

aphedges: @sigmavirus24, I did point out in the PR description that this is not strictly related. I can move it to another PR if you want.

While I agree that `tox` simplifies things, I have similar concerns to @Mr-Pepe. When I made this PR, I was working on improving type annotations to make the code easier to understand. With the current `tox` configuration, there isn't a way to run just `mypy` by itself, and even if there were, there wouldn't be an easy workflow to trying different arguments.

I agree that moving as many options to a single place would make it easier for alternative development workflows to be done. `tox` would be the ultimate test in CI and ideally before making a PR, but for fast iteration during development, it's overkill.

aphedges: > we should have some other way to run `black` and `isort`

Running [`make format`](https://github.com/PyCQA/pydocstyle/blob/5a1c4d3a128ce662301a80532a578f37f8d5c3c3/Makefile#L3) is an existing alternative way to run `black` and `isort` that is already part of the repository.

samj1912 (approved):

Issue #537, opened 2021-05-13
Add flag to ignore some functions, classes or class methods


takos22:
## My problem
I document my class initialization inside the class docstring, therefore pydocstyle complains about undocumented `__init__` methods. 
## My solution idea
It would be great to have the ability to have a flag (for example ``--ignore-func``) that would work for both functions, classes and class methods with maybe a regex. So for ignoring the `__init__` the flag could be ``--ignore-func='.*\.__init__'``, with `.*` for the class name, `\.` to say that it's a class method and `__init__` to ignore that method.

KaleabTessera:
Would be a great feature!

Issue #515, opened 2020-10-22
Don't force to add docstrings to __new__


facundobatista:
This should be considered a magic method (controlled by `D105`), or at least have a different error code (which may be `D107` together with `__init__` or even a new one).

DetachHead:
same with `__call__`

PR #587, opened 2022-04-04
Add the ability to skip checking some functions.

mgilson:
There are times when you might not want to check a function's
docstring.  Specifically, we frequently want to ignore
test files.  There is already an exemption that causes test
functions to be considered non-public.  This extends on
that idea and makes it configurable so a user can cause
a test function to be allowed to have no docstring.

This is a potential implementation for #537, #515 

Thanks for submitting a PR!


src/pydocstyle/checker.py
```diff
@@ -150,9 +151,19 @@ def check_source(
                         len(ignore_decorators.findall(dec.name)) > 0
                         for dec in definition.decorators
                     )
+                    # Only skip checking the docstring if absent.
+                    # If the docstring is present, then check it as normal
+                    name_skip = (
+                        ignore_functions is not None
+                        and not definition.docstring
+                        and bool(ignore_functions.findall(definition.name))
```
aphedges, line 159: `search()` is probably a better option than `findall()` because it only needs to find one instance of the pattern instead of all. However, function names should short enough that the performance difference is negligible, so I'm not sure a change is needed.

aphedges (commented):
From looking at the code itself, this looks good except for a small concern. I'll try it out later this week to see if it works well.

psycoder21 (approved):

Issue #459, opened 2020-03-07
[Bug] D417 not raised for missing parameter when docstring contains further sections


nioncode:
The following example does not raise a D417:
```python
def foo(a, b):
    """Foobar.

    Args:
        a: test

    Raises:
        KeyError: description
    """
    pass
```
There is also no error raised when replacing `Raises` with `Returns`, i.e. D417 is broken for both.

Without the `Raises` section, everything works correctly, i.e.:
```python
def foo(a, b):
    """Foobar.

    Args:
        a: test

    """
    pass
```
prints `D417: Missing argument descriptions in the docstring (argument(s) b are missing descriptions in 'foo' docstring)`.

samj1912:
Yeah - the issue is related to how we treat conventions as a bundle of error codes, instead of a first class entity. If we indeed take advantage of the fact that the user is specifying the convention type - we can avoid such ambigious cases. @Nurdok what are your thoughts on this? Also currently any google style docstring that has a returns section yields no errors for other sections since its detected as a numpy section.

Nurdok:
@samj1912 - as I commented on #478, I think using the `--convention` flag to determine how to parse docstrings SGTM, but I'd like to see a more detailed proposal of how this would work.

adamjstewart:
@samj1912 if I understand correctly, any docstring containing a section that is present in both Google-style and Numpy-style, such as:

* Returns
* Raises
* Yields
* See Also
* Notes
* References
* Examples
* Attributes
* Methods

will be interpreted as Numpy-style instead of Google-style, is that correct? These are obviously very common sections for docstrings, so this effectively means that auto-detection of Google-style docstrings doesn't currently work.

There are many possible fixes for this:

1. Instead of checking for numpy section names, check for numpy-only section names
2. Use the `--convention` flag to force a particular style
3. Add a new flag to force a particular style and keep `--convention` as a bundle of error codes only

Do you have a particular preference for any of these? I think 1 is the quickest/easiest and most flexible since it would work most of the time, even if a single project mixes both Numpy and Google style in different methods. I'm also perfectly okay with 2 or 3 since my projects don't mix styles and I can easily specify which convention I use.

I'm happy to submit a PR to try to implement any of these options, although I may need pointers on which files contain the style autodetection code.

Mr-Pepe:
@adamjstewart The file `checker.py` contains the following, starting from line 1062:

```Python
@check_for(Definition)
[...]

PR #589, opened 2022-04-14
Introduce convention class

Mr-Pepe:
This PR implements the first step towards using the specified convention to assume the style of docstrings and not rely on figuring out whether a docstring uses NumPy or Google style (#459). 

This PR introduces a new class `Convention` that holds the error codes to check. It also keeps the name of the convention to use. I imagine that an instance of this class should be passed to the `ConventionChecker`. Convention-specific checks like `_check_numpy_sections` and `_check_google_sections` can then be executed based on the specified convention. This would eventually solve #459.

I added tests for the code I added and all existing tests still pass.

### DIFF
--- a/src/pydocstyle/conventions.py
+++ b/src/pydocstyle/conventions.py
@@ -1,0 +1,109 @@
+"""This module contains the convention definitions."""
+
+
+from typing import Literal, Set
+
+from pydocstyle.violations import all_errors
+
+CONVENTION_NAMES = ("pep257", "numpy", "google")
+
+
+convention_errors = {
+    'pep257': all_errors
+    - {
+        'D203',
+        'D212',
+        'D213',
+        'D214',
+        'D215',
+        'D404',
+        'D405',
+        'D406',
+        'D407',
+        'D408',
+        'D409',
+        'D410',
+        'D411',
+        'D413',
+        'D415',
+        'D416',
+        'D417',
+        'D418',
+    },
+    'numpy': all_errors
+    - {
+        'D107',
+        'D203',
+        'D212',
+        'D213',
+        'D402',
+        'D413',
+        'D415',
+        'D416',
+        'D417',
+    },
+    'google': all_errors
+    - {
+        'D203',
+        'D204',
+        'D213',
+        'D215',
+        'D400',
+        'D401',
+        'D404',
+        'D406',
+        'D407',
+        'D408',
+        'D409',
+        'D413',
+    },
+}
+
+
+class Convention:
+    """This class defines the convention to use for checking docstrings."""
+
+    def __init__(
+        self, name: Literal["pep257", "numpy", "google"] = "pep257"
+    ) -> None:
+        """Initialize the convention.
+
+        The convention has two purposes. First, it holds the error codes to be
+        checked. Second, it defines how to treat docstrings, eliminating the
+        need for extra logic to determine whether a docstring is a NumPy or
+        Google style docstring.
+
+        Each convention has a set of error codes to check as a baseline.
+        Specific error codes can be added or removed via the
+        :code:`add_error_codes` or :codes:`remove_error_codes` methods.
+
+        Args:
+            name (Literal["pep257", "numpy", "google"], optional): The convention to use. Defaults to "pep257".
+
+        Raises:
+            ValueError: _description_
+        """
+        if name not in CONVENTION_NAMES:
+            raise ValueError(
+                f"Convention '{name}' is invalid. Must be one "
+                f"of {CONVENTION_NAMES}."
+            )
+
+        self.name = name
+        self.error_codes = convention_errors[name]
+
+    def add_error_codes(self, error_codes: Set[str]) -> None:
+        """Add additional error codes to the convention.
+
+        Args:
+            error_codes (Set[str]): The error codes to also check.
+        """
+        self.error_codes = self.error_codes.union(error_codes)
+
+    def remove_error_codes(self, error_codes: Set[str]) -> None:
+        """Remove error codes from the convention.
+
+        Args:
+            error_codes (Set[str]): The error codes to ignore.
+        """
+        self.error_codes = self.error_codes - error_codes
--- a/src/pydocstyle/conventions.py
+++ b/src/pydocstyle/conventions.py
@@ -84,10 +84,7 @@ class Convention:
             ValueError: _description_
         """
         if name not in CONVENTION_NAMES:
-            raise ValueError(
-                f"Convention '{name}' is invalid. Must be one "
-                f"of {CONVENTION_NAMES}."
-            )
+            name = "pep257"
 
         self.name = name
         self.error_codes = convention_errors[name]
--- a/src/pydocstyle/__init__.py
+++ b/src/pydocstyle/__init__.py
@@ -2,4 +2,4 @@
 from .checker import ConventionChecker, check
 from .parser import AllError
 from .utils import __version__
-from .violations import Error, conventions
+from .violations import Error
--- a/src/pydocstyle/config.py
+++ b/src/pydocstyle/config.py
@@ -10,8 +10,9 @@ from configparser import NoOptionError, NoSectionError, RawConfigParser
 from functools import reduce
 from re import compile as re
 
+from .conventions import CONVENTION_NAMES, Convention
 from .utils import __version__, log
-from .violations import ErrorRegistry, conventions
+from .violations import ErrorRegistry
 
 try:
     import toml
@@ -189,7 +190,7 @@ class ConfigurationParser:
     DEFAULT_PROPERTY_DECORATORS = (
         "property,cached_property,functools.cached_property"
     )
-    DEFAULT_CONVENTION = conventions.pep257
+    DEFAULT_CONVENTION = Convention()
 
     PROJECT_CONFIG_FILES = (
         'setup.cfg',
@@ -597,7 +598,7 @@ class ConfigurationParser:
         elif options.select is not None:
             checked_codes = cls._expand_error_codes(options.select)
         elif options.convention is not None:
-            checked_codes = getattr(conventions, options.convention)
+            checked_codes = Convention(options.convention).error_codes
 
         # To not override the conventions nor the options - copy them.
         return copy.deepcopy(checked_codes)
@@ -642,7 +643,7 @@ class ConfigurationParser:
         """Extract the codes needed to be checked from `options`."""
         checked_codes = cls._get_exclusive_error_codes(options)
         if checked_codes is None:
-            checked_codes = cls.DEFAULT_CONVENTION
+            checked_codes = cls.DEFAULT_CONVENTION.error_codes
 
         cls._set_add_options(checked_codes, options)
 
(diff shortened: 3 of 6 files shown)


Mr-Pepe:
Is there any interest in moving this forward?

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,107 @@
...
+        'D400',
+        'D401',
+        'D404',
+        'D406',
+        'D407',
+        'D408',
+        'D409',
+        'D413',
+    },
+}
```
benji-york, line 61: I'm wondering if there is a common subset of the errors being removed from each of the conventions that makes sense to name.  E.g., something along the lines of "very strict rules".

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(
+        self, name: Literal["pep257", "numpy", "google"] = "pep257"
+    ) -> None:
+        """Initialize the convention.
+
+        The convention has two purposes. First, it holds the error codes to be
+        checked. Second, it defines how to treat docstrings, eliminating the
+        need for extra logic to determine whether a docstring is a NumPy or
+        Google style docstring.
```
benji-york, line 75: The second part seems especially useful to me.  Going very far down the road to DWIM ("do what I mean") gets pretty painful.  Users end up having to trick the tool into doing what they want instead of just specifying it as this approach allows.

benji-york (approved):
I'm not a pydocstyle maintainer, so this review doesn't mean much, but this looks like a nice improvement to the code.  It mostly just ties together ideas that were pre-existing into concrete representations.

I made a couple of small comments inline.

samj1912:
@adamjstewart / @Pierre-Sassoulas would you like to review this?

requirements/runtime.txt
```diff
@@ -1,2 +1,3 @@
 snowballstemmer==1.2.1
 toml==0.10.2
+typing_extensions==4.1.1
```
Pierre-Sassoulas, line 3: Not sure for what this is used, I don't see it in the pyproject.toml, and also it was already like this before... But ``typing_extensions`` is used by a lot of packages, the constraint should be more forgiving as to be compatible with more of them imo (using >= instead of == at least).

Pierre-Sassoulas (commented):
Isn't the numpy/google conventions also following pep 257 ?

adamjstewart:
On vacation this week but will review next week.

adamjstewart (commented):
Alright, finally got a chance to test this. Using this branch, I was able to find and fix several new bugs in [TorchGeo](https://github.com/microsoft/torchgeo)'s docs: https://github.com/microsoft/torchgeo/pull/1011. I can confirm that this PR fixes #459!

Note that there were a couple of odd things I noticed. First was the output of pydocstyle:
```console
$ pydocstyle .
./torchgeo/datasets/gid15.py:247 in public method `plot`:
        D417: Missing argument descriptions in the docstring (argument(s) suptitle are missing descriptions in 'plot' docstring)
./torchgeo/datasets/cyclone.py:226 in public method `plot`:
        D417: Missing argument descriptions in the docstring (argument(s) show_titles, suptitle are missing descriptions in 'plot' docstring)
./torchgeo/datamodules/vaihingen.py:39 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) num_patches_per_tile, num_workers, patch_size, val_split_pct are missing descriptions in '__init__' docstring)
./torchgeo/datamodules/deepglobelandcover.py:37 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) num_patches_per_tile, num_workers, patch_size, val_split_pct are missing descriptions in '__init__' docstring)
./torchgeo/datamodules/potsdam.py:39 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) num_patches_per_tile, num_workers, patch_size, val_split_pct are missing descriptions in '__init__' docstring)
./torchgeo/models/rcf.py:36 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) bias, features, kernel_size, seed are missing descriptions in '__init__' docstring)
./torchgeo/samplers/single.py:76 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) length, roi, size, units are missing descriptions in '__init__' docstring)
./torchgeo/samplers/single.py:184 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) roi, size, stride, units are missing descriptions in '__init__' docstring)
./torchgeo/samplers/single.py:287 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) roi, shuffle are missing descriptions in '__init__' docstring)
./torchgeo/samplers/batch.py:74 in public method `__init__`:
        D417: Missing argument descriptions in the docstring (argument(s) batch_size, length, roi, size, units are missing descriptions in '__init__' docstring)
```
None of these functions are missing any argument descriptions, the problems lay elsewhere. So the error messages don't seem correct.

The other odd thing I noticed was that pydocstyle complains if you put Sphinx directives like `versionadded` after Args/Returns. I don't believe Google's style guide precludes this, so I'm surprised to see pydocstyle complain.

Both of these issues feel like bugs in pydocstyle, not bugs in this PR. I think we should first push forward with this PR and I can open issues for the above bugs at a later date.

requirements/runtime.txt
adamjstewart: This dep is needed for `typing.Literal`, which wasn't introduced until Python 3.8. So this dependency is actually needed. Dependencies in `requirements/*` can and should actually be pinned to a specific version (mostly for CI purposes, we should manage these files with dependabot). For users actually installing pydocstyle, these files aren't used, the deps listed in `pyproject.toml` or `setup.py` are. Before merging, we should rebase this PR on master and add `typing_extensions` to `pyproject.toml`. They don't have a changelog so I don't know what version of typing_extensions added Literal though. Although honestly, we could prob just type things as a string instead of as a Literal. That's probably simpler.

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,107 @@
...
+        'D413',
+    },
+}
+
+
+class Convention:
+    """This class defines the convention to use for checking docstrings."""
+
+    def __init__(
+        self, name: Literal["pep257", "numpy", "google"] = "pep257"
```
adamjstewart, line 68: We should prob just type this as a `str` instead of a `Literal`, then we don't need typing_extensions. The point of [PEP 586](https://peps.python.org/pep-0586/) (which introduced `Literal`) was to support functions that change return type based on which literal value is passed as input. Since we don't need that here, the type checker doesn't really benefit from using Literal over str.

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,107 @@
...
+        checked. Second, it defines how to treat docstrings, eliminating the
+        need for extra logic to determine whether a docstring is a NumPy or
+        Google style docstring.
+
+        Each convention has a set of error codes to check as a baseline.
+        Specific error codes can be added or removed via the
+        :code:`add_error_codes` or :codes:`remove_error_codes` methods.
+
+        Args:
+            name (Literal["pep257", "numpy", "google"], optional): The convention to use. Defaults to "pep257".
```
adamjstewart, line 82: No need to put the type in the docstring too, [sphinx.ext.autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints) will automatically pull the types from the function signature and use those in the docs.

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Raises:
+            ValueError: _description_
+        """
+        if name not in CONVENTION_NAMES:
+            name = "pep257"
```
adamjstewart, line 88: This function doesn't raise a ValueError at the moment, but I think it actually should. Using an undefined convention name and silently replacing it with pep257 without at least a warning is not a good idea. It seems like pep257 is used by default if users don't specify anything in their `pyproject.toml`, so raising an error here will still maintain backwards compatibility.

src/pydocstyle/conventions.py
adamjstewart: I actually don't love the "all_errors - {subset}" logic, but not just because of code duplication. Any time anyone adds a new convention, we have to add it to this list. Personally, it makes more sense to me to make these additive instead of subtractive. But let's save this for another issue/PR since this PR doesn't change anything, just moves it around.

src/pydocstyle/conventions.py
Mr-Pepe: I like using `Literal` when the number of options is limited because static analysis will save me when I make a typo. However, I like your proposal of raising an error at runtime instead :+1:

Mr-Pepe:
Rebased on top of master and addressed feedback

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,105 @@
+"""This module contains the convention definitions."""
+
+from typing import Set
+
+from typing_extensions import Literal
```
adamjstewart, line 5: This import can be removed now

adamjstewart (commented):
I think something went wrong with the rebase, this branch no longer catches the mistakes found previously in https://github.com/microsoft/torchgeo/pull/1011 and no longer fixes #459.

adamjstewart:
We should add a test case for #459. All of our tests are passing but #459 is still broken.

src/pydocstyle/conventions.py
Mr-Pepe: Removed. The tests did not crash because `importlib-metadata` pulled in `typing_extensions` as a dependency.

Mr-Pepe:
I'm actually surprised that this PR fixed #459 for you because it is not supposed to, as outlined in the PR description.  I did not touch anything regarding how docstrings are treated or how checks are executed so there must have been something else going on during your tests. This PR only formalizes the concept of conventions a bit.

adamjstewart:
Gotcha, yeah I have no idea why things were working before the rebase, maybe because the branch was old? I'm fine with formalizing the concept and then fixing #459 in a separate PR. Thanks for your work on this!

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,103 @@
...
+        'D406',
+        'D407',
+        'D408',
+        'D409',
+        'D413',
+    },
+}
+
+
+class Convention:
```
adamjstewart, line 62: What's the advantage of using a class instead of just using an Enum and keeping the AttrDict?

src/pydocstyle/conventions.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def add_error_codes(self, error_codes: Set[str]) -> None:
+        """Add additional error codes to the convention.
+
+        Args:
+            error_codes: The error codes to also check.
+        """
+        self.error_codes = self.error_codes.union(error_codes)
+
+    def remove_error_codes(self, error_codes: Set[str]) -> None:
```
adamjstewart, line 97: If you instead define `__add__`/`__or__` and `__sub__`, you can run:
```python
convention = Convention()
convention += {"D103"}
convention -= {"D101", "D102"}
```

adamjstewart (commented):
Not sure how well I'll be able to review this since I'm not that familiar with the internals. I guess I'm not sure why this PR is needed as a precursor to fix #459, which is what I really care about.

src/pydocstyle/conventions.py
Mr-Pepe: My idea was to eventually have the convention class itself check the code and select the appropriate checker function. You can probably also just throw the name of the convention into the convention checker and that's it.

src/pydocstyle/conventions.py
Mr-Pepe: Yes that's possible, but at least to me it makes the whole thing less obvious, without offering a real benefit.

PR #597, opened 2022-06-27
add `--max-error-percentage` flags

mashanz:
Thanks for submitting a PR!

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -1080,6 +1080,45 @@ class ConventionChecker:
 
 parse = Parser()
 
+def error_file(filenames,
+    select=None,
+    ignore=None,
+    ignore_decorators=None,
+    property_decorators=None,
+    ignore_inline_noqa=False,
+):
+    if select is not None and ignore is not None:
+        raise IllegalConfiguration(
+            'Cannot pass both select and ignore. '
+            'They are mutually exclusive.'
+        )
+    elif select is not None:
+        checked_codes = select
+    elif ignore is not None:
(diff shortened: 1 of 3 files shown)


sigmavirus24:
What issue is this related to? Why would someone want to set a percentage of failures in order to succeed? The code is not clear on what it's trying to do and I don't see any prior discussion of this

mashanz:
Hi @sigmavirus24 ,
There is rule on my company to create documentation on software that we are building, and we check the documentation coverage using pydocstyle and running it on CI/CD pipeline. 

but because our app is to huge but less documented, we set at least 50% the file is fully documented. and we will gradually decrease the error percentage until the software 100% fully documented overtime.

because `pydocstyle` didn't have those use case to cover this specific needs, we introduce `--max-error-percentage` on `pydocstyle` to cover this use case.

I hope this clarify this PR. Thanks.

sigmavirus24:
It does clarify, thank you. I would structure this differently, personally and even then I'm not convinced it belongs in this tool. If it were to live here, it should look more like coverage.py and its flags.

mashanz:
> 

Thanks for the advice. Will update this PR.

sigmavirus24:
Actually, I wouldn't even integrate that here because it's already covered by https://github.com/econchick/interrogate

mashanz:
> Actually, I wouldn't even integrate that here because it's already covered by https://github.com/econchick/interrogate

oh, That package is new to me, thanks @sigmavirus24 for the information. It's very helpful.
I think we can close this PR and I will use that tools on our CI/CD's

PR #607, opened 2022-09-13
Allow to run with all Python optimizations enabled

### LINKED ISSUE #606: pycodestyle is not compatible with `PYTHONOPTIMIZE=2`
pycodestyle [relies on function doctrings](https://github.com/PyCQA/pydocstyle/blob/a6fe4222012e990042c86cdaaa904a8d059752ee/src/pydocstyle/checker.py#L167) to show the linting errors. When you run a program with `PYTHONOPTIMIZE=2` environment variable or [the `-OO` option](https://docs.python.org/3/using/cmdline.html#cmdoption-OO) the docstrings are stripped, so the next error is raised:

```
... (traceback)

  File "/.../site-packages/pydocstyle/checker.py", line 160, in check_source
    partition = this_check.__doc__.partition('.\n')
AttributeError: 'NoneType' object has no attribute 'partition'
```

mondeja:
Resolves #606

### DIFF
--- a/src/pydocstyle/checker.py
+++ b/src/pydocstyle/checker.py
@@ -2,6 +2,7 @@
 
 import ast
 import string
+import sys
 import tokenize as tk
 from collections import namedtuple
 from itertools import chain, takewhile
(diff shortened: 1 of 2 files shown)


sigmavirus24:
I'm not speaking as a maintainer here but I don't think this is desirable in the slightest. Using the docstrings for explaining the error is maybe questionable but just removing the docstrings altogether is a big mistake and hurts maintainability and readability of this tool

mondeja:
I'm open to duplicate the explanations to avoid the docstrings stripping and add a test to synchronize docstrings with explanations. If a maintainer can take a final decission I'll be happy to implement it 👍🏼

Issue #111, opened 2015-03-12
Files passed as arguments ignored by --match


jpulec:
It looks like when you pass specific filenames to pep257, it ignores if you have set the ```--match``` option.

For example:

```
pep257 foo.py bar.py --match='(?!foo).*\.py'
```
 Will still check foo.py and bar.py. While I understand why this makes sense, since there is a default match pattern that ignores 'test_.*', and you want to be able to easily use pep257 on a file prefixed with 'test_.*' if desired, iit means that pep257 doesn't perform as one might expect when using something like xargs to pass a bunch of files into it.

jpulec:
PR in #112

jayvdb:
Close?

jpulec:
Is this seen as a non-standard use case, or that it breaks expected behavior?

Nurdok:
I'm not sure how I feel about this. On the one hand, I get your point, but I also feel that passing an explicit file name should take precedence over the matching regex. Anyone else cares to pitch in?

erichiller:
Was this reverted? The behavior I am seeing shows that an explicitly passed in file will override `match`

aaaaahaaaaa:
I'm noticing it indeed overrides `match` and `match_dir` configs when using the pre-commit hook since the files are passed as positional arguments. Which makes the hook useless as far as I can tell because it checks files which are meant to be ignored.

A workaround is to use a `local` hook to run the shell command directly:
```
- repo: local
  hooks:
    - id: pydocstyle
      name: pydocstyle
      types: [python]
      entry: pydocstyle
      language: system
      pass_filenames: false
```

danmoz:
> On the one hand, I get your point, but I also feel that passing an explicit file name should take precedence over the matching regex. Anyone else cares to pitch in?

Currently the only way to exclude certain files is using negative lookahead patterns like `--match='(?!test_).*\.py'`. Since this is ignored when files are passed in (i.e. when using pre-commit) there is no way to exclude files passed in on the CLI.

The only thing pre-commit users can do is use `pass_filenames: false` (as suggested in previous comment) which causes pydocstyle to run on the whole codebase. But in a large codebase, this is too slow.

IMHO, for the pre-commit support to be useful, there needs to be a way to exclude certain file patterns, even if they are passed in explicitly on the CLI.

What if we introduced a switch that prevents `match` from being ignored when filenames are passed in? That would retain backwards compatibility, but bring joy back into the life of pydocstyle/pre-commit users.

[...]

PR #610, opened 2022-11-07
fix: do not pass file names with pre-commit

yajo:
This is the only way pydocstyle will obey its own match rules.

Thanks for submitting a PR!


samj1912 (approved):

bszonye:
This change is causing pre-commit to emit a warning:
```
$ pre-commit run --all-files
[WARNING] Unexpected key(s) present on https://github.com/pycqa/pydocstyle => pydocstyle: pass_file_names
```

bszonye:
Also, this appears to change the documented behavior here:
https://www.pydocstyle.org/en/stable/usage.html?highlight=pre-commit#usage-with-the-pre-commit-git-hooks-framework

> Checked-in python files will be passed as positional arguments so no need to use `--match=*.py`. You can also use command line arguments instead of configuration files to achieve the same effect with less files.

If that's changing then the documentation needs an update and possibly instructions for how to get the old behavior.

bszonye:
Should I move my comments here to a new issue? I noticed that a revert commit has been opened and closed a couple times now.

samj1912:
This commit has been reverted.

bszonye:
Thanks! By the way, I discovered accidentally that the underlying issue that motivated this PR might have already been addressed. Please see my comment here:
https://github.com/PyCQA/pydocstyle/issues/111#issuecomment-1374921227

bszonye:
Aha, I was still seeing the error in pre-commit because this:
```
- repo: https://github.com/pycqa/pydocstyle
  rev: 6.2.3
  hooks:
  - id: pydocstyle
```
gets the broken config from the 6.2.3 release. Reverting my config to 6.2.2 works around the problem.

PR #653, opened 2023-09-13
Fix crash when using f-strings

### LINKED ISSUE #640: Crash with simple f-string
Pydocstyle crashes with the below code:

Code: `f"bar {123}"`
Traceback:
```
  File "venv/bin/pydocstyle", line 8, in <module>
    sys.exit(main())
  File "venv/lib/python3.7/site-packages/pydocstyle/cli.py", line 79, in main
    sys.exit(run_pydocstyle())
  File "venv/lib/python3.7/site-packages/pydocstyle/cli.py", line 54, in run_pydocstyle
    ignore_self_only_init=ignore_self_only_init,
  File "venv/lib/python3.7/site-packages/pydocstyle/checker.py", line 1171, in check
    ignore_self_only_init,
  File "venv/lib/python3.7/site-packages/pydocstyle/checker.py", line 159, in check_source
    self, definition, definition.docstring
  File "venv/lib/python3.7/site-packages/pydocstyle/checker.py", line 244, in check_docstring_empty
    if docstring and is_blank(ast.literal_eval(docstring)):
  File "/root/.pyenv/versions/3.7.2/lib/python3.7/ast.py", line 91, in literal_eval
    return _convert(node_or_string)
  File "/root/.pyenv/versions/3.7.2/lib/python3.7/ast.py", line 90, in _convert
    return _convert_signed_num(node)
  File "/root/.pyenv/versions/3.7.2/lib/python3.7/ast.py", line 63, in _convert_signed_num
    return _convert_num(node)
  File "/root/.pyenv/versions/3.7.2/lib/python3.7/ast.py", line 55, in _convert_num
    raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: <_ast.JoinedStr object at 0x7f296ff35a20>
```

Python 3.7.2
Pydocstyle: 6.3.0

kasium:
Closes #640

Thanks for submitting a PR!

-  [X] Add unit tests and integration tests where applicable.  
   

sigmavirus24:
This prevents the crash but it misleads people and will lead to big reports of "but there's a docstring there" because it doesn't inform people that the f string is invalid and cannot be a docstring. I would not accept this PR as is

kasium:
@sigmavirus24 Reason for that would be that in https://github.com/PyCQA/pydocstyle/pull/381 the following was suggested

> Mistakenly using an f-string for a docstring does not cause a syntax error, so we shouldn't raise a ParseError here.
Instead, I suggest we rename this as eval_docstring and just return an empty string if we find an f-string.

sigmavirus24:
The idea for adding a new error code was not dropped. The PR lost steam due to people being busy. The new error code is correct behavior. How we achieve that is up for debate

PR #657, opened 2023-11-01
Add readthedocs configuration

parona-source:
* Configuration file is mandatory after readthedocs 10.8.0.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,13 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+sphinx:
+  configuration: docs/conf.py
+
+python:
+  install:
+    - requirements: requirements/docs.txt


parona-source:
Fixes CI in https://github.com/PyCQA/pydocstyle/pull/656

parona-source:
Tried to match the configuration in .github/workflows, but readthedocs doesn't support ubuntu-latest as an option so I had to pick a specific ubuntu release.

docs/release_notes.rst
```diff
@@ -4,6 +4,14 @@ Release Notes
 **pydocstyle** version numbers follow the
 `Semantic Versioning <http://semver.org/>`_ specification.
 
+
+Current development version
+---------------------------
+
+Bug Fixes
+
+* Add readthedocs configuration to fix GitHub action (#657).
```
sigmavirus24, line 13: This isn't a bug fix that affects users in any meaningful way. We don't need it in the changelog

sigmavirus24 (changes requested):

sigmavirus24 (approved):

